PR #4, opened 2019-01-07
Fixes to animation system

eanders-MS:
Developing the chess app exposed a few animation bugs:
1. Actor transform was not getting sent to the app when animating. Transforms should not be synchronized to other clients when animating, but the app should still get updates if you've subscribed to transform changes on the actor.
2. ResetAnimation handler wasn't implemented on the client.
3. Animation enable/disable is now controlled by setting the animation speed to 1 or 0. This allows us to set the animation time even when the animation isn't playing. This was necessary for ResetAnimation to work.

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
@@ -516,10 +516,11 @@ namespace MixedRealityExtension.Core
                 return false;
             }
 
-            // Sync all updates to the app if we're operating in a peer-authoritative model and we're the authoritative peer, unless the actor is currently animating.
+            // Sync all updates to the app if we're operating in a peer-authoritative model and we're the authoritative peer.
             if (App.OperatingModel == OperatingModel.PeerAuthoritative && App.IsAuthoritativePeer)
             {
-                return !Animating;
+                // return !Animating;
+                return true;
             }
 
             // Sync this update to the app if we're operating in a server-authoritative model and the app has registered interest in this setting.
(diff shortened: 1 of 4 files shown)


stevenvergenz:
Retarget PR against dev?

MREUnityRuntime/MREUnityRuntimeLib/Core/Components/AnimationComponent.cs
```diff
@@ -182,14 +188,14 @@ internal void ResetAnimation(string animationName)
             {
                 if (animation[animationName] != null)
                 {
-                    if (animation[animationName].enabled)
+                    if (animation[animationName].speed > 0)
```
stevenvergenz: >speed > 0 [](start = 49, length = 9)

Is it possible to play an animation backward with a negative speed?

eanders-MS: Good catch. We don't support this in the API currently, but it might be something we expose later.

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -516,10 +516,10 @@ private bool ShouldSync(SubscriptionType subscriptions, SubscriptionType flag)
                 return false;
             }
 
-            // Sync all updates to the app if we're operating in a peer-authoritative model and we're the authoritative peer, unless the actor is currently animating.
+            // Sync all updates to the app if we're operating in a peer-authoritative model and we're the authoritative peer.
             if (App.OperatingModel == OperatingModel.PeerAuthoritative && App.IsAuthoritativePeer)
             {
-                return !Animating;
+                return true;
```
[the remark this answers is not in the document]
eanders-MS: Maybe... it would be interesting to diff the generated IL. I was erring on the side of clarity.

eanders-MS:
Didn't adequately account for root motion and animation blending. Abandoning and going back to the drawing board.

PR #5, opened 2019-01-10
Actor load error handling

stevenvergenz:
* Update UnityGLTF to latest [upstream PR](https://github.com/KhronosGroup/UnityGLTF/pull/343)
* Change CreateFromLibrary, CreateFromGLTF, and LoadAssetsFromGLTF to be `async`. All other CreateFrom functions are now synchronous. No callbacks anymore! This was necessary to correctly propagate exceptions thrown from the loader code.
* Removed the glTF loading queue and related infrastructure, as it was all being run synchronously anyway. Tested with the new `gltf-concurrent-test` to make sure they don't need to be serialized, and it all seems to work.
* Explicitly prevent message handlers from throwing. Our reflection goo doesn't handle it well, so better to process it in the individual handlers.


MREUnityRuntime/MREUnityRuntimeLib/Util/UtilMethods.cs
```diff
@@ -96,5 +96,21 @@ public static void GetUrlParts(string url, out string rootUrl, out string filena
 
             filename = uri.Segments[uri.Segments.Length - 1];
         }
+
+        public static string FormatException(Exception ex)
```
eanders-MS: Make internal unless you need to call it from Altspace.

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -381,88 +381,64 @@ internal bool OwnsActor(IActor actor)
 
         #region Methods - Private
 
-        private void ProcessCreateFromGLTFQueue()
+        [CommandHandler(typeof(CreateFromGLTF))]
+        private async Task OnCreateFromGLTF(CreateFromGLTF payload)
         {
-            // GLTFLoader can only load one model at a time :(
-            if (Monitor.TryEnter(_createFromGLTFQueue))
+            if (_actorManager.HasActor(payload.Actor?.Id) && _actorManager.IsActorReserved(payload.Actor?.Id))
```
eanders-MS: I think these conditions should be OR'd, not AND'd:
```ts
if (_actorManager.HasActor(payload.Actor?.Id) || _actorManager.IsActorReserved(payload.Actor?.Id))
```
(I realize this is my code from a while back :) )

MREUnityRuntime/MREUnityRuntimeLib/Util/UtilMethods.cs
eanders-MS: Sending a stack trace back to the app is gross. Let's just send the "unexpected error" message.

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
stevenvergenz: Is there a reason we succeeded in this case? My PR changes this to a failure, but I want to make sure that's the right behavior.

---
In reply to: [246835616](https://github.com/Microsoft/mixed-reality-extension-unity/pull/5#discussion_r246835616) [](ancestors = 246835616)

MREUnityRuntime/MREUnityRuntimeLib/Util/UtilMethods.cs
stevenvergenz: My desired flow here is that we send back enough data for the app developer to file a github issue that's actionable. This seems like the best way for us. If the message were just "model load error", we'd have to repro the error to troubleshoot, which slows things down.

---
In reply to: [246836428](https://github.com/Microsoft/mixed-reality-extension-unity/pull/5#discussion_r246836428) [](ancestors = 246836428)

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
eanders-MS: Let's try going with failure and see how that goes.

PR #8, opened 2019-01-16
[Unity] Actor patching improvements

eanders-MS:
An optimization on the Node side exposed an unsafe assumption in the client's handling of RPC messages.

Also: Sped up walk speed in the Functional Test scene.


eanders-MS:
Releasing a new npm package doesn't have to wait for this change to go out in the DLL. This bug only affects rpc calls received by the client that have no parameters, which doesn't affect existing apps. The functional test will complain in the Unity log, but doesn't break the tests themselves.

PR #12, opened 2019-01-19
Rewrite test layout logic

stevenvergenz:


sorenhan:
If this is regarding the idea of possibly changing which way all text objects face, you should hold off on this. I want to discuss this in more detail with Eric, before we decide on any changes here.

eanders-MS:
This change is solely regarding the layout of the functional tests in the Unity scene. Is that correct @stevenvergenz?

stevenvergenz:
Correct. Unfortunately that means that all text in the functional tests faces away from you, which is a little awkward. I still think this PR is useful though, as it fixes the layout logic.

PR #15, opened 2019-01-25
Fix AssetSource deserialization

stevenvergenz:


MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetSource.cs
```diff
@@ -5,26 +5,29 @@
...
         /// The type of container the asset came from.
         /// </summary>
-        public AssetContainerType ContainerType { get; }
+        public AssetContainerType ContainerType { get; set; }
 
         /// <summary>
         /// The URL of the asset's container.
         /// </summary>
-        public Uri Uri { get; }
+        public Uri Uri { get; set; }
```
eanders-MS: This needs to be a string. If the URL is malformed it will fail the JSON serialization (Uri class will fail to parse it). That's too early to be performing field validation.

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetSource.cs
```diff
@@ -5,26 +5,35 @@
...
         /// The URL of the asset's container.
         /// </summary>
-        public Uri Uri { get; }
+        public string Uri { get; set; }
+
+        private Uri parsedUri;
+        /// <summary>
+        /// The parsed URI of the asset's container.
+        /// </summary>
+        public Uri ParsedUri => parsedUri = parsedUri ?? new Uri(Uri);
```
eanders-MS: >ParsedUri [](start = 19, length = 9)

Please make sure this field is excluded from JSON serialization. Because this class is part of a network payload, it is unsafe to assume it will never be sent to the app. I'd recommend not having this field and just parsing the url in LoadAssetsFromGLTF call.

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetSource.cs
```diff
@@ -5,26 +5,35 @@
...
 
         /// <summary>
         /// The location of the asset within the container. Type-dependent.
         /// </summary>
-        public string InternalId { get; }
+        public string InternalId { get; set; }
```
eanders-MS: >InternalId [](start = 22, length = 10)

Exclude this field from JSON serialization as well?

stevenvergenz: No, this field is supported on both sides. It's unused right now, but I think we'll need it when I implement UnloadAsset.

---
In reply to: [251068885](https://github.com/Microsoft/mixed-reality-extension-unity/pull/15#discussion_r251068885) [](ancestors = 251068885)

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetSource.cs
stevenvergenz: It makes sense to be able to cache the Uri result. I've excluded it.

---
In reply to: [251068734](https://github.com/Microsoft/mixed-reality-extension-unity/pull/15#discussion_r251068734) [](ancestors = 251068734)

PR #18, opened 2019-01-31
[Unity] New version compatibility scheme

eanders-MS:
C# side of the new version compatibility check in PR https://github.com/Microsoft/mixed-reality-extension-sdk/pull/157

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Constants.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Constants.cs
@@ -28,8 +28,16 @@ namespace MixedRealityExtension
          */
         internal const string SessionHeader = "x-ms-mixed-reality-extension-sessionid";
         internal const string PlatformHeader = "x-ms-mixed-reality-extension-platformid";
+        internal const string LegacyProtocolVersionHeader = "x-ms-mixed-reality-extension-protocol-version";
         internal const string CurrentClientVersionHeader = "x-ms-mixed-reality-extension-client-version";
         internal const string MinimumSupportedSDKVersionHeader = "x-ms-mixed-reality-extension-min-sdk-version";
+
+        /*
+         * Legacy Protocol Version - Left over from the old protocol versioning scheme.
+         * Keeping in for now to allow a smoother transition to the new system.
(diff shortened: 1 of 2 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Constants.cs
```diff
@@ -23,15 +23,26 @@ static Constants()
             SerializerSettings.Converters.Add(new PayloadConverter());
         }
 
+        /*
+         * Headers passed in MRE connect requests.
+         */
         internal const string SessionHeader = "x-ms-mixed-reality-extension-sessionid";
         internal const string PlatformHeader = "x-ms-mixed-reality-extension-platformid";
-        internal const string ProtocolVersionHeader = "x-ms-mixed-reality-extension-protocol-version";
```
stevenvergenz: >ProtocolVersionHeader [](start = 30, length = 21)

We can't stop sending these without breaking old SDK versions. We'll have to make sure to message the community about this. Alternatively, can we keep sending protocol versions too for a couple weeks? That would prevent random breakage.

MREUnityRuntime/MREUnityRuntimeLib/Constants.cs
```diff
@@ -23,15 +23,26 @@ static Constants()
...
         /*
-         * PROTOCOL VERSION
-         * This is the version of the wire protocol this host library speaks. This value is sent to the MRE app when
-         * connecting. If client's protocol version is unsupported, the connection will be rejected.
+         * Minimum Supported SDK version
+         * The oldest SDK version that runs. Since compatibility with older MREs is essential, changing
+         * this *is* a big deal, and requires discussion and signoff from the dev team.
          */
-        internal const int ProtocolVersion = 1;
+        internal const string MinimumSupportedSDKVersion = "0.5";
```
stevenvergenz: >"0.5" [](start = 59, length = 5)

Won't this be 0.6 or later? The minimum supported version has to be the first SDK version that's actually checking these headers, right?

sorenhan:
Approved for Red branch

PR #19, opened 2019-02-01
[Unity] Texture asset support

stevenvergenz:
* Add Vector2 class, patch, and extensions
* Fill out the new Texture asset type, with resolution and wrap settings
* Add a texture, offset, and scale to the Material asset
* Fix bugs with patching materials


MREUnityRuntime/MREUnityRuntimeLib/Patching/Types/VectorPatch.cs
```diff
@@ -1,69 +0,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
-using MixedRealityExtension.Core.Types;
-using System;
-using UnityEngine;
-
-namespace MixedRealityExtension.Patching.Types
-{
-    public class Vector3Patch : IEquatable<Vector3Patch>, IPatchable
```
eanders-MS: >Vector3Patch [](start = 17, length = 12)

Was this class unused?

PR #20, opened 2019-02-02
Fix reparenting

eanders-MS:
Steven noticed that reparenting wasn't working. This fixes that. FYI I coded it to preserve the child's local transform when reparenting.

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
@@ -155,7 +155,8 @@ namespace MixedRealityExtension.Core
 
         internal void ApplyPatch(ActorPatch actorPatch)
         {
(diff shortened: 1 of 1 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -155,7 +155,8 @@ internal void SynchronizeApp()
 
         internal void ApplyPatch(ActorPatch actorPatch)
         {
-            var parent = ParentId != null ? App.FindActor(ParentId.Value) : Parent;
+            var parentId = actorPatch.ParentId ?? ParentId;
+            var parent = parentId != null ? App.FindActor(parentId.Value) : Parent;
```
stevenvergenz: Why are we patching parent in a separate function than the rest of the properties? Is this a holdover from the previous architecture?

PR #22, opened 2019-02-05
[Unity] Create materials and textures directly

stevenvergenz:
* Add support for a new `create-asset` message, which generates a single standalone asset
* Add a public default material initialization parameter, which is used for primitives, and should be used for glTFs in Altspace.
* Added a string Uri field to Texture objects
* Added a handy `dictionary.GetOrCreate(key, defaultConstructor)` utility function

### DIFF
--- a/MRETestBed/Assets/TestBed Assets/Scripts/MREComponent.cs
+++ b/MRETestBed/Assets/TestBed Assets/Scripts/MREComponent.cs
@@ -88,9 +88,9 @@ public class MREComponent : MonoBehaviour
         if (!_apiInitialized)
         {
             MREAPI.InitializeAPI(
+                DefaultPrimMaterial,
                 behaviorFactory: new BehaviorFactory(),
                 textFactory: new MWTextFactory(SerifFont, SansSerifFont),
-                primitiveFactory: new MWPrimitiveFactory(DefaultPrimMaterial),
                 libraryFactory: new ResourceFactory(),
                 assetCache: new AssetCache(new GameObject("MRE Asset Cache")),
                 logger: new MRELogger());
--- a/MREUnityRuntime/MREUnityRuntimeLib/API/MREApi.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/API/MREApi.cs
@@ -43,7 +43,7 @@ namespace MixedRealityExtension.API
             AppsAPI.DefaultMaterial = defaultMaterial;
             AppsAPI.BehaviorFactory = behaviorFactory;
             AppsAPI.TextFactory = textFactory ?? throw new ArgumentException($"{nameof(textFactory)} cannot be null");
-            AppsAPI.PrimitiveFactory = primitiveFactory;
+            AppsAPI.PrimitiveFactory = primitiveFactory ?? new MWPrimitiveFactory();
             AppsAPI.LibraryResourceFactory = libraryFactory;
             AppsAPI.AssetCache = assetCache ?? new AssetCache();
             AppsAPI.GLTFImporterFactory = gltfImporterFactory ?? new GLTFImporterFactory();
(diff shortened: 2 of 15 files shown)


MRETestBed/Assets/TestBed Assets/Scripts/MREComponent.cs
```diff
@@ -88,9 +88,9 @@ void Start()
         if (!_apiInitialized)
         {
             MREAPI.InitializeAPI(
+                DefaultPrimMaterial,
                 behaviorFactory: new BehaviorFactory(),
                 textFactory: new MWTextFactory(SerifFont, SansSerifFont),
-                primitiveFactory: new MWPrimitiveFactory(DefaultPrimMaterial),
```
eanders-MS: >primitiveFactory [](start = 16, length = 16)

Did you intend to remove primitiveFactory?

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -336,7 +336,7 @@ public IUser FindUser(Guid id)
 
         internal void OnReceive(Message message)
         {
-            if (message.Payload is LoadAssets || message.Payload is AssetUpdate)
+            if (message.Payload is LoadAssets || message.Payload is CreateAsset || message.Payload is AssetUpdate)
```
eanders-MS: This pattern isn't scaling well. Will have to find a better way eventually.

MRETestBed/Assets/TestBed Assets/Scripts/MREComponent.cs
stevenvergenz: The initialization in the Testbed, yes. The constructor no longer takes arguments, so it's redundant with the default value.

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
stevenvergenz: Agreed. If I have to add one more message, I'll make a types array.

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetCache.cs
```diff
@@ -41,7 +49,7 @@ public GameObject EmptyTemplate()
         }
 
         /// <inheritdoc cref="GetAssetIdsInSource"/>
-        public IEnumerable<Guid> GetAssetIdsInSource(AssetSource source)
+        public IEnumerable<Guid> GetAssetIdsInSource(AssetSource source = null)
```
eanders-MS: >null [](start = 74, length = 4)

I'm pretty certain TryGetValue will throw if source is null.

MREUnityRuntime/MREUnityRuntimeLib/API/MREApi.cs
```diff
@@ -30,6 +31,7 @@ public static class MREAPI
         /// <param name="gltfImporterFactory">The glTF loader factory. Uses default GLTFSceneImporter if omitted.</param>
         /// <param name="logger">The logger to be used by the MRE SDK.</param>
         public static void InitializeAPI(
+            UnityEngine.Material defaultMaterial,
             IBehaviorFactory behaviorFactory = null,
             ITextFactory textFactory = null,
             IPrimitiveFactory primitiveFactory = null,
```
eanders-MS: >IPrimitiveFactory primitiveFactory = null, [](start = 12, length = 42)

Do you need to remove primitiveFactory here?

stevenvergenz: I'm leaving this in so that host apps can supply their own implementations of the PrimitiveFactory. They don't have to now though.

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetCache.cs
stevenvergenz: You are correct, good catch!

PR #26, opened 2019-02-08
Support material alpha blending and cutout

stevenvergenz:
* Adds `AlphaMode` and `AlphaCutoff` fields to the material patch
* Abstracts Unity material reads and writes to an `IMaterialPatcher` plugin instance. This is necessary because Unity's API does not provide a shader-agnostic transparency mode accessor.

### DIFF
--- a/MRETestBed/Assets/Scenes/FunctionalTest-localhost.unity
+++ b/MRETestBed/Assets/Scenes/FunctionalTest-localhost.unity
@@ -851,8 +851,9 @@ MonoBehaviour:
     type: 2}
   MREURL: ws://localhost:3901
   testNames:
-  - asset-early-assignment-test
+  - asset-early-assignment
   - asset-preload-test
+  - mutable-asset-test
   - clock-sync-test
   - gltf-animation-test
   - gltf-concurrency-test
@@ -861,7 +862,6 @@ MonoBehaviour:
   - interpolation-test
   - library-test
   - look-at-test
-  - mutable-asset-test
   - primitives-test
   - reparent-test
   - rigid-body-test
--- a/MRETestBed/Assets/Scenes/FunctionalTest-localhost.unity
+++ b/MRETestBed/Assets/Scenes/FunctionalTest-localhost.unity
@@ -851,9 +851,9 @@ MonoBehaviour:
     type: 2}
   MREURL: ws://localhost:3901
   testNames:
-  - asset-early-assignment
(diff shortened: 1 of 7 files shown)


MRETestBed/Assets/Scenes/FunctionalTest-localhost.unity
```diff
@@ -851,8 +851,9 @@ MonoBehaviour:
     type: 2}
   MREURL: ws://localhost:3901
   testNames:
-  - asset-early-assignment-test
+  - asset-early-assignment
```
eanders-MS: >asset-early-assignment [](start = 4, length = 22)

Please have each test end in 'test', to match the test names in app.ts.

MRETestBed/Assets/Scenes/FunctionalTest-localhost.unity
```diff
@@ -851,8 +851,9 @@ MonoBehaviour:
...
   - asset-preload-test
+  - mutable-asset-test
```
eanders-MS: >mutable-asset-test [](start = 4, length = 18)

Let's keep this list alphabetized.

stevenvergenz: Can I rename the test then? I'd like to have all the asset tests together.

MRETestBed/Assets/Scenes/FunctionalTest-localhost.unity
stevenvergenz: I just noticed that app.ts actually has it without test too. I can rename that instead if we want.

MRETestBed/Assets/Scenes/FunctionalTest-localhost.unity
eanders-MS: Certainly!

MRETestBed/Assets/Scenes/FunctionalTest-localhost.unity
eanders-MS: Let's have every test end in '-test'.

MRETestBed/Assets/Scenes/FunctionalTest-localhost.unity
stevenvergenz: Okay, I've updated these to what they "should" be, but they'll be wrong until I change the names SDK-side.

PR #28, opened 2019-02-13
[Unity] Fixes for stability

eanders-MS:
Found and fixed numerous issues exposed by stressing network traffic in different ways.
- Fix patching of rigid body position and rotation (happens via transform sync, but applied to the rigid body directly).
- Rigid body patch was applying new values when none were present in the patch.  Added protection for this.
- Updated DashFormattedEnumConverter to work with Nullable enum types.
- Updated GeneratePatch functions to return null if there was no patch, enabling better actor-update deltas by not sending empty objects.
- In the case of a failure to process a network message, send a reply message indicating the failure. This will unblock any promises on the app side, and also communicate some detail about the failure.

Also:
- General cleanup of actor patching code.

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Patching/PatchingUtils.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Patching/PatchingUtils.cs
@@ -40,6 +40,26 @@ namespace MixedRealityExtension.Patching
             }
         }
 
+        public static string GeneratePatch(string _old, string _new)
+        {
+            if (_old == null && _new != null)
+            {
+                return _new;
+            }
+            else if (_new == null)
+            {
+                return null;
+            }
+            if (_old != _new)
+            {
+                return _new;
+            }
+            else
+            {
+                return null;
+            }
+        }
+
         public static Vector3Patch GeneratePatch(MWVector3 _old, Vector3 _new)
         {
             if (_old == null && _new != null)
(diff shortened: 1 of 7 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Patching/PatchingUtils.cs
```diff
@@ -40,6 +40,26 @@ internal static class PatchingUtilMethods
...
+        public static string GeneratePatch(string _old, string _new)
+        {
+            if (_old == null && _new != null)
+            {
+                return _new;
+            }
+            else if (_new == null)
+            {
+                return null;
+            }
```
sorenhan: Couldn't this just be if(_old == null) {return _new;}  - whether _new is null or not doesn't matter, we just return new

sorenhan: or rather if(_old == null || _new == null || _old != _new) return _new;?


---
In reply to: [256591338](https://github.com/Microsoft/mixed-reality-extension-unity/pull/28#discussion_r256591338) [](ancestors = 256591338)

eanders-MS: Yes, but I was following the pattern used elsewhere in the file.

---
In reply to: [256592130](https://github.com/Microsoft/mixed-reality-extension-unity/pull/28#discussion_r256592130) [](ancestors = 256592130,256591338)

PR #31, opened 2019-02-15
[Unity] Remove unused code

eanders-MS:
Removed the notion of user subscriptions and user transform. We weren't using them, and probably never will.


eanders-MS:
I've updated the PRs to leave the user-update message.

PR #34, opened 2019-02-25
[Unity] Avatar attachments

eanders-MS:
#### Overview
- A new method on IUserInfo called GetAttachPoint(name) returns a Unity transform representing the attach point. Actor is reparented from the app scene root to this transform.

- Actors attached to avatars are not synchronized to other clients. They're attached to the user's avatar independently on each client, obviating the need to synchronize them.

#### Other changes
- Renamed SubscriptionType -> ActorComponentType
- IUserInfo.LookAtPosition now returns a nullable type. This is a temporary thing. LookAtPosition will be removed entirely in the near future.
- Tweak to WebSocket dispose.
- Log json of message if we fail handling it for any reason.


MREUnityRuntime/MREUnityRuntimeLib/Core/Attachment.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Threading.Tasks;
+
+namespace MixedRealityExtension.Core
+{
+    public class Attachment : IEquatable<Attachment>
+    {
+        internal string AttachPoint { get; set; }
+        internal Guid UserId { get; set; }
+
+        public bool Equals(Attachment other)
```
tombuMS: For IEquatable we should also override the Object.Equals and Object.GetHashCode methods as a convention.  BehaviorHandlerBase.cs has a great example of this.

MREUnityRuntime/MREUnityRuntimeLib/Messaging/Payloads/NetworkCommandPayloads.cs
```diff
@@ -231,14 +231,14 @@ public class UpdateSubscriptions : NetworkCommandPayload
         public Guid Id { get; set; }
 
         /// <summary>
-        /// The subscription types to add to the object. See <see cref="SubscriptionType"/>.
+        /// The subscription types to add to the object. See <see cref="ActorComponentType"/>.
```
tombuMS: Might want to update these to "The actor component type to subscribe/unsubscribe to."

MREUnityRuntime/MREUnityRuntimeLib/Messaging/Payloads/Payloads.cs
```diff
@@ -33,27 +33,47 @@ public enum OperationResultCode
     /// The flags that are subscription types for message patching notifications.
```
tombuMS: Update description

MREUnityRuntime/MREUnityRuntimeLib/Patching/PatchingUtils.cs
```diff
@@ -61,6 +62,16 @@ public static string GeneratePatch(string _old, string _new)
         }
 
         public static Vector3Patch GeneratePatch(MWVector3 _old, Vector3 _new)
+        {
+            MWVector3 _mwNew = null;
+            if (_new != null)
+            {
+                _mwNew = _new.ToMWVector3();
```
tombuMS: The methods ToMW*() all allocate a new MW type, and it looks like we are just doing this here to have an object to use in the comparison to generate a patch.  We should look at an alternative way to do this so that we don't have a ton of allocations just to generate patches.  See the below methods as well.

eanders-MS: Reverted this change. Was no longer used.

---
In reply to: [259984496](https://github.com/Microsoft/mixed-reality-extension-unity/pull/34#discussion_r259984496) [](ancestors = 259984496)

MREUnityRuntime/MREUnityRuntimeLib/Core/Attachment.cs
eanders-MS: Added a comment stating this class is not suitable for use as a hash key or dictionary key. I'm also adding an explicit override of GetHashCode in order to ward off future red flags here.

---
In reply to: [259979640](https://github.com/Microsoft/mixed-reality-extension-unity/pull/34#discussion_r259979640) [](ancestors = 259979640)

tombuMS: Make sure you add the override for the Object.Equals method. So that object equality matches that of IEquatable.

PR #36, opened 2019-02-27
Set collider API integration support for Unity

tombuMS:
Added in the backing support on Unity for setting and resetting a collider on an actor.  There is also logic adding in the collider to the initial actor patch upon creation for the case that the newly created actor already has a collider of type box, sphere or mesh.  This makes the app aware that an actor has an existing collider on an actor in the case it was there at the time of creation.

**_Note: Contains a slight cleanup effort around the IsPatched methods for our patching system._**:

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
@@ -550,6 +550,7 @@ namespace MixedRealityExtension.Core
                     // We have a collider already of the same type as the desired new geometry.
                     // Update its values instead of removing and adding a new one.
                     colliderGeometry.Patch(_collider);
+                    return Collider;
                 }
                 else
                 {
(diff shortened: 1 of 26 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -522,6 +533,56 @@ private RigidBody AddRigidBody()
...
+                    Collider = null;
+                }
+            }
+
+            UnityCollider unityCollider = null;
+
+            switch (colliderType)
+            {
+                case ColliderType.Box:
+                    var boxCollider = gameObject.AddComponent<BoxCollider>();
```
eanders-MS: >var boxCollider = gameObject.AddComponent<BoxCollider>(); [](start = 20, length = 57)

If _collider wasn't destroyed above, would calling SetCollider again result in multiple Collider components on the actor?

tombuMS: Yes it would.  Colliders are just components on the game object, and you can technically have as many colliders that you want on the game object.  They may not work as expected, but it is allowed by Unity.

tombuMS: I see what you mean.  Good catch.  Updated to return the collider above in the case that we just patch it since it is already the correct type.

PR #39, opened 2019-03-01
[Unity] LookAt, take 3

eanders-MS:
Implements LookAtComponent.
Also cleaned up a few build warnings (non-XML).


MREUnityRuntime/MREUnityRuntimeLib/Util/Unity/AssetFetcher.cs
```diff
@@ -8,9 +8,9 @@ namespace MixedRealityExtension.Util.Unity
 
     public static class AssetFetcher<T> where T : class
     {
-        public struct FetchResult<T>
+        public struct FetchResult<TT> where TT : T
```
stevenvergenz: >TT [](start = 34, length = 2)

Did you mean to check this in? This seems unrelated to LookAt.

eanders-MS: Fixing a build warning: T shadowed T from parent class.

PR #40, opened 2019-03-01
fix sync for looping sounds after the first playthrough has completed

sorenhan:
Unity quirk: if an audiosource time is larger than the audioclip length, it doesn't automatically calculate the proper time in the middle of the loop. This works around that quirk

This also doesn't create a soundinstance if it's a non-looping sound that would have ended already - it would have played once, even though it was supposed to not play at all-

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
@@ -1027,17 +1027,25 @@ namespace MixedRealityExtension.App
                     var audioClip = MREAPI.AppsAPI.AssetCache.GetAsset(payload.SoundAssetId) as AudioClip;
                     if (audioClip != null)
                     {
-                        var soundInstance = actor.gameObject.AddComponent<AudioSource>();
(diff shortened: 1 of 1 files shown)


MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -1027,17 +1027,25 @@ private void OnSetSoundState(SetSoundState payload)
...
-                        soundInstance.minDistance = 1.0f;
-                        soundInstance.maxDistance = 1000000.0f;
-                        ApplySoundStateOptions(soundInstance, payload.Options);
-                        soundInstance.Play();
-                        _soundInstances.Add(payload.Id, soundInstance);
-                        _unpausedSoundInstances.Add(payload.Id);
+                        float offset = payload.StartTimeOffset;
+                        if (payload.Options.Looping != null && payload.Options.Looping.Value)
+                        {
+                            offset = payload.StartTimeOffset%audioClip.length;
```
tombuMS: Please leave spaces between operators.  Makes this difficult to read when looking through code quickly.

PR #41, opened 2019-03-01
refactored pause/resume as not commands

sorenhan:


MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -1041,10 +1064,14 @@ private void OnSetSoundState(SetSoundState payload)
                             soundInstance.spread = 90.0f;   //only affects multichannel sounds. Default to 50% spread, 50% stereo.
                             soundInstance.minDistance = 1.0f;
                             soundInstance.maxDistance = 1000000.0f;
-                            ApplySoundStateOptions(soundInstance, payload.Options);
-                            soundInstance.Play();
-                            _soundInstances.Add(payload.Id, soundInstance);
                             _unpausedSoundInstances.Add(payload.Id);
```
stevenvergenz: > _unpausedSoundInstances.Add(payload.Id); [](start = 27, length = 41)

Shouldn't this only be added if the instance is unpaused? There's no check.

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -943,12 +943,25 @@ private void OnSetAnimationState(SetAnimationState payload)
 
 
 
+        private static Dictionary<Guid, AudioSource> _soundInstances = new Dictionary<Guid, AudioSource>();
+        private static List<Guid> _unpausedSoundInstances = new List<Guid>();
+        private int _soundStoppedCheckIndex = 0;
```
stevenvergenz: Normally I would say that these need to go at the top of the class, but nobody wants to scroll past 900 lines to get to them. This class is getting too big.

PR #42, opened 2019-03-01
Add some attach points to MRETestBed character

eanders-MS:


MRETestBed/Assets/TestBed Assets/Scripts/UserInfo.cs
```diff
@@ -28,8 +28,32 @@ public UserInfo(Guid id)
     public event MWEventHandler BeforeAvatarDestroyed;
     public event MWEventHandler AfterAvatarCreated;
 
+    private static Transform FindChildRecursive(Transform parent, string name)
+    {
```
tombuMS: This is a useful helper function and should be in an appropriate file in the Util folder and namespace.  Not private in this class I would think.

MRETestBed/Assets/TestBed Assets/Scripts/UserInfo.cs
```diff
@@ -28,8 +28,32 @@ public UserInfo(Guid id)
...
+        Transform transform = parent.Find(name);
+        if (transform != null)
+        {
+            return transform;
+        }
+        for (int i = 0; i < parent.childCount; ++i)
+        {
```
tombuMS: This can be a ranged for loop by doing `foreach (Transform t in parent.GetComponentsInChildren<Transform>())` which is cleaner.  Ranged for is always preferred over indexed for loop to range safety.

stevenvergenz: For that matter, this whole method could be a one-liner: `parent.GetComponentsInChildren<Transform>().FirstOrDefault(t => t.name == name)`

---
In reply to: [261803994](https://github.com/Microsoft/mixed-reality-extension-unity/pull/42#discussion_r261803994) [](ancestors = 261803994)

MRETestBed/Assets/TestBed Assets/Scripts/UserInfo.cs
stevenvergenz: I don't know how often this method would actually be used, as we're generally not searching for Transforms, but I agree that Utils is the right place for this.

---
In reply to: [261803456](https://github.com/Microsoft/mixed-reality-extension-unity/pull/42#discussion_r261803456) [](ancestors = 261803456)

eanders-MS: No need to move this to the UnityRuntime project unless we have a need for it there.

eanders-MS: I like the one-liner though :)

MRETestBed/Assets/TestBed Assets/Scripts/UserInfo.cs
tombuMS: Yeah.  That too.

evhan55:
Hi @eanders-MS ,

I am using the MRETestBed for some tests on the HoloLens1. I am using Actors with attachments at 'head' and it works well in Altspace, but not in the MRETestBed. This commit makes it look like the MRETestBed does attach the 'head' somehow via the 'socket-head' GameObject.... is that understanding correct?  Thank you!

eanders-MS:
That correct, this PR adds a GameObject for the 'socket-head' attach point. In what way is it not working?

evhan55:
@eanders-MS Thank you for confirming!

In my MRE, I attach an actor at the attach point `'head'`, and then I subscribe to the actor's transforms.
For Altspace users, the actor shows up and updates its position correctly whether seen in AltspaceVR or in the HoloLens1. (it's so cool!)
However, for HoloLens1 users, the actor shows up at position 0,0,0 and never updates position correctly, whether seen in AltspaceVR or in the HoloLens1.

So, it seems to me like the actor attaches correctly when users join via Altspace, but the actor does not attach correctly when users join via the HoloLens1. I am using the MRETestBed as my base in Unity, with minimal changes to it.

evhan55:
@eanders-MS oh and please let me know if it's better to submit this as an issue!

eanders-MS:
Sounds like a possible regression in the testbed. Would you mind opening an issue over at https://github.com/microsoft/mixed-reality-extension-sdk/issues? That way we can track it and it will have more visibility. I'm amazed you found this PR :)

evhan55:
Hi @eanders-MS 

I had asked about this problem over in the MRE SDK Discord and @stevenvergenz mentioned that the Unity app should be handling the attachments. And since I was using the MRETestBed Unity app I thought to come to this repo and do a search on 'attach' to see if the attachment points were handled in a way I was expecting. That led me to this PR!

Thank you again, and I'll submit a new issue over in the sdk repo!

PR #44, opened 2019-03-06
fix sound unity exceptions on actor destroy

sorenhan:
when an actor is destroyed, unity's gameobject destroy automatically destroys all audiosources. we need to clean up our lists tracking these instances. This is not fast, but functional.


eanders-MS:
Integrated via my other PR

PR #45, opened 2019-03-06
[Unity] Move message queuing to the client

eanders-MS:
The client can now handle messages for `Actor`s that haven't been created yet. It does this by creating a message queue for every actor and accumulating messages there. Once the actor is created, it will dispatch the messages in serial to the actor. All actor messaging goes through this mechanism now. This means that messages are processed serially per actor. This is a new constraint, but I think it will be acceptable since we have so few async actor messages, mainly create-animation.

One architectural change is that every message handler takes an onCompleteCallback parameter. Message handlers must call this exactly once and within one minute of the message being dispatched. There is new code in `CommandManager` to enforce this.

These changes are pretty extensive. Apologies for the burden.

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
@@ -166,17 +166,11 @@ namespace MixedRealityExtension.Core
 
         internal void ExecuteRigidBodyCommands(RigidBodyCommands commandPayload, Action onCompleteCallback)
         {
-            try
+            foreach (var command in commandPayload.CommandPayloads.OfType<ICommandPayload>())
             {
-                foreach (var command in commandPayload.CommandPayloads.OfType<ICommandPayload>())
-                {
-                    App.ExecuteCommandPayload(this, command, null);
-                }
+                App.ExecuteCommandPayload(this, command, null);
             }
-            finally
-            {
-                onCompleteCallback?.Invoke();
-            }
+            onCompleteCallback?.Invoke();
         }
 
         internal void Destroy()
@@ -861,42 +855,24 @@ namespace MixedRealityExtension.Core
         [CommandHandler(typeof(ActorCorrection))]
         private void OnActorCorrection(ActorCorrection payload, Action onCompleteCallback)
         {
-            try
-            {
-                // TODO: Interpolate this change onto the actor.
-                SynchronizeEngine(payload.Actor);
-            }
-            finally
-            {
-                onCompleteCallback?.Invoke();
-            }
+            // TODO: Interpolate this change onto the actor.
+            SynchronizeEngine(payload.Actor);
+            onCompleteCallback?.Invoke();
         }
 
         [CommandHandler(typeof(ActorUpdate))]
         private void OnActorUpdate(ActorUpdate payload, Action onCompleteCallback)
         {
-            try
-            {
-                SynchronizeEngine(payload.Actor);
-            }
-            finally
-            {
-                onCompleteCallback?.Invoke();
-            }
+            SynchronizeEngine(payload.Actor);
+            onCompleteCallback?.Invoke();
         }
 
         [CommandHandler(typeof(UpdateSubscriptions))]
         private void OnUpdateSubscriptions(UpdateSubscriptions payload, Action onCompleteCallback)
         {
-            try
-            {
-                RemoveSubscriptions(payload.Removes);
-                AddSubscriptions(payload.Adds);
-            }
-            finally
-            {
-                onCompleteCallback?.Invoke();
-            }
+            RemoveSubscriptions(payload.Removes);
+            AddSubscriptions(payload.Adds);
+            onCompleteCallback?.Invoke();
         }
 
         [CommandHandler(typeof(RigidBodyCommands))]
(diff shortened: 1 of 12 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Core/SoundManager.cs
```diff
@@ -0,0 +1,142 @@
...
+            if (_soundStoppedCheckIndex >= _unpausedSoundInstances.Count)
+            {
+                _soundStoppedCheckIndex = 0;
+            }
+            else
+            {
+                var id = _unpausedSoundInstances[_soundStoppedCheckIndex];
+                AudioSource soundInstance;
+                if (_soundInstances.TryGetValue(id, out soundInstance) && !soundInstance.isPlaying)
+                {
```
tombuMS: Use a scoped variable here for soundInstance as `out AudioSource soundInstance`

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -847,44 +856,378 @@ private bool CanSync()
...
+                                soundInstance.Play();
+                            }
+                            App.SoundManager.AddSoundInstance(payload.Id, soundInstance);
+                        }
+                    }
+                }
+                else
+                {
+                    AudioSource soundInstance;
+                    if (App.SoundManager.TryGetSoundInstance(payload.Id, out soundInstance))
```
tombuMS: Scoped variable for soundInstance `out AudioSource soundInstance`

MREUnityRuntime/MREUnityRuntimeLib/Core/ActorCommandQueue.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            if (actor == null)
+            {
+                actor = app.FindActor(actorId) as Actor;
+            }
+
+            if (actor != null && queue.Count > 0)
+            {
+                activeCommand = queue.Dequeue();
+                app.ExecuteCommandPayload(actor, activeCommand.Payload, OnCommandComplete);
```
tombuMS: Look at handling exceptions here instead of forcing each handler function to have a try/finally block.  Centralizes our exception handling to one point for the command queue processing.

MREUnityRuntime/MREUnityRuntimeLib/Messaging/Commands/CommandManager.cs
```diff
@@ -10,36 +10,169 @@ namespace MixedRealityExtension.Messaging.Commands
...
+                    while (_pendingCompletionCallbacks.Count > 0)
+                    {
+                        // Callbacks are naturally sorted by creation time.
+                        var callback = _pendingCompletionCallbacks.First();
+                        if (callback.CreationTime <= currTime + QueuedCompletionCallbackTimeout)
+                        {
+                            // Command handler did not invoke its onCompleteCallback in a timely manner.
+                            MREAPI.Logger.LogError($"ERROR: Timeout waiting for onCompleteCallback invocation from {callback.Location}");
+                            _pendingCompletionCallbacks.Remove(callback);
+                            // PONDER: Should we forceably invoke the callback from here?
```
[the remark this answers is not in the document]
eanders-MS: If this command was executed from ActorCommandQueue.Update then the queue would stall and never process another message.

---
In reply to: [263082536](https://github.com/Microsoft/mixed-reality-extension-unity/pull/45#discussion_r263082536) [](ancestors = 263082536)

MREUnityRuntime/MREUnityRuntimeLib/Core/SoundManager.cs
eanders-MS: Not my code, but sure. Was copy/pasted from App.

---
In reply to: [263092709](https://github.com/Microsoft/mixed-reality-extension-unity/pull/45#discussion_r263092709) [](ancestors = 263092709)

MREUnityRuntime/MREUnityRuntimeLib/Core/ActorManager.cs
```diff
@@ -2,19 +2,23 @@
...
 namespace MixedRealityExtension.Core
 {
-    internal class ActorManager
+    internal class ActorManager : ICommandHandlerContext
     {
         private MixedRealityExtensionApp _app;
         private Dictionary<Guid, Actor> _actorMapping = new Dictionary<Guid, Actor>();
-        private HashSet<Guid> _reservations = new HashSet<Guid>();
+        private Dictionary<Guid, ActorCommandQueue> _actorCommandQueues = new Dictionary<Guid, ActorCommandQueue>();
+        private List<Action> _uponStable = new List<Action>();
```
eanders-MS: >_uponStable [](start = 29, length = 11)

Note: This is only needed due to the way we sync animations currently. Once that is refactored this can go away. We're going to change it to work like sound does.

MREUnityRuntime/MREUnityRuntimeLib/Messaging/Commands/Command.cs
```diff
@@ -1,23 +1,26 @@
...
+        private Action _onCompleteCallback;
 
-        public Command(T commandPayload)
+        public Command(T commandPayload, Action onCompleteCallback)
         {
             _commandPayload = commandPayload;
+            _onCompleteCallback = onCompleteCallback;
         }
 
         public void Execute(ICommandHandlerContext handlerContext, MethodInfo handlerMethod)
```
eanders-MS: >void [](start = 15, length = 4)

Question: Instead of an onCompleteCallback, what if this returned a Task that could be awaited?

MREUnityRuntime/MREUnityRuntimeLib/Messaging/Commands/CommandManager.cs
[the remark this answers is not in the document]
tombuMS: How should we handle this case?  The PONDER here signals we should think about whether to do this, but it sounds like something is fatal if we don't based on your last comment.  Am I missing something?

MREUnityRuntime/MREUnityRuntimeLib/Messaging/Commands/Command.cs
tombuMS: Is it necessary to have this wrapped in to a task rather than using the mechanism you have already built?  There is overhead associated with async methods and Tasks.  I would say that for the use case here, we should probably not be using tasks as the overhead is unnecessary.

MREUnityRuntime/MREUnityRuntimeLib/Messaging/Commands/CommandManager.cs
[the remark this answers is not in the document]
eanders-MS: If we get here then we have a logic bug in a handler. I think the error msg is sufficient, and it names the specific handler so that's helpful.

---
In reply to: [263142163](https://github.com/Microsoft/mixed-reality-extension-unity/pull/45#discussion_r263142163) [](ancestors = 263142163)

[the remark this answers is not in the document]
tombuMS: Sounds good.

PR #48, opened 2019-03-07
Process actor destroy via command queue

eanders-MS:
This ensures that actors are created before destroyed.

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Core/ActorManager.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Core/ActorManager.cs
@@ -37,35 +37,43 @@ namespace MixedRealityExtension.Core
             return actor;
         }
 
-        internal void DestroyActors(IEnumerable<Guid> ids, IList<Trace> traces)
+        internal void DestroyActors(IEnumerable<Guid> ids)
         {
             foreach (var id in ids)
             {
-                if (!_actorMapping.ContainsKey(id))
+                var destroyActorCommand = new LocalCommand
                 {
-                    var message = "destroy-actors: Actor not found: " + id.ToString() + ".";
-                    if (traces != null)
+                    Command = () =>
                     {
-                        traces.Add(new Trace()
+                        if (_actorCommandQueues.TryGetValue(id, out ActorCommandQueue queue))
                         {
-                            Severity = TraceSeverity.Warning,
-                            Message = message
-                        });
-                    }
+                            // Clear the queue so that pending messages are canceled.
+                            queue.Clear();
+                            _actorCommandQueues.Remove(id);
+                        }
 
-                    MREAPI.Logger.LogError(message);
-                }
-                else
-                {
-                    var actor = _actorMapping[id];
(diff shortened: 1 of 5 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Messaging/Payloads/PayloadTypeRegistry.cs
```diff
@@ -61,6 +61,7 @@ namespace MixedRealityExtension.Messaging.Payloads
     [PayloadType(typeof(AssetsLoaded), "assets-loaded")]
     [PayloadType(typeof(AssetUpdate), "asset-update")]
     [PayloadType(typeof(CreateAsset), "create-asset")]
+    [PayloadType(typeof(LocalCommand), "local-command")]
```
eanders-MS: @tombuMS Is this needed if the payload is never sent or received?

MREUnityRuntime/MREUnityRuntimeLib/Core/ActorManager.cs
```diff
@@ -101,10 +109,16 @@ internal void ProcessActorCommand(Guid actorId, NetworkCommandPayload payload, A
             queue.Enqueue(payload, onCompleteCallback);
         }
 
+        private List<ActorCommandQueue> _queuesForUpdate = new List<ActorCommandQueue>();
+
```
sorenhan: Why is this a class member, instead of just being inside update()? It seems to be only used inside update(), and cleared at start of each update?

sorenhan:
>         internal void ProcessActorCommand(Guid actorId, NetworkCommandPayload payload, Action onCompleteCallback)

wouldn't it be better to name it QueueActorCommand? it doesn't actually process right away

---
Refers to: MREUnityRuntime/MREUnityRuntimeLib/Core/ActorManager.cs:102 in 56ef02d. [](commit_id = 56ef02d119b68d50d1e875231e2c313d89a46d4b, deletion_comment = False)

MREUnityRuntime/MREUnityRuntimeLib/Core/ActorManager.cs
eanders-MS: An optimization. I didn't want to reallocate it every frame.

---
In reply to: [263543869](https://github.com/Microsoft/mixed-reality-extension-unity/pull/48#discussion_r263543869) [](ancestors = 263543869)

eanders-MS:
>         internal void ProcessActorCommand(Guid actorId, NetworkCommandPayload payload, Action onCompleteCallback)

I considered that. I felt that would be leaking an implementation detail.

---
In reply to: [470671951](https://github.com/Microsoft/mixed-reality-extension-unity/pull/48#issuecomment-470671951) [](ancestors = 470671951)

---
Refers to: MREUnityRuntime/MREUnityRuntimeLib/Core/ActorManager.cs:102 in 56ef02d. [](commit_id = 56ef02d119b68d50d1e875231e2c313d89a46d4b, deletion_comment = False)

PR #59, opened 2019-03-15
Fix parenting to pending actors

stevenvergenz:
* Wait until actor creation to attach children
* Add transform-test to the FT testbed scene
* Add a library cube to the testbed
* Fix the testbed HelloWorld URL


eanders-MS:
We need to be able to set parentId before the said parent object is loaded. Does this change support this? Definitely an oversight in my earlier changes...

eanders-MS:
Oh, I see the queued command. Nice.

---
In reply to: [473386312](https://github.com/Microsoft/mixed-reality-extension-unity/pull/59#issuecomment-473386312) [](ancestors = 473386312)

PR #60, opened 2019-03-21
[Unity] User Masking

stevenvergenz:
* Change `appearance.enabled` to `UInt32`
* Add `UInt32 IUser.Groups`
* Evaluate actor visibility against user group mask, re-evaluate when the user group mask changes
* Accept `user-update` messages from apps


MREUnityRuntime/MREUnityRuntimeLib/Core/User.cs
```diff
@@ -18,6 +20,8 @@ internal class User : MixedRealityExtensionObject, IUser
 
         public IUserInfo UserInfo { get; private set; }
 
+        public UInt32 Groups { get; internal set; } = 1;
```
eanders-MS: What is the special meaning of group 1?

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -79,12 +79,13 @@ public new string Name
...
 
         internal Guid MaterialId { get; set; } = Guid.Empty;
-
+        
         internal bool Grabbable { get; private set; }
         
-        private bool AppearanceEnabled { get; set; } = true;
-
-        private bool ActiveAndEnabled => ((Parent as Actor)?.ActiveAndEnabled ?? true) && AppearanceEnabled;
+        private UInt32 AppearanceEnabled = UInt32.MaxValue;
```
eanders-MS: cslint: private members should start with a lowercase letter : `AppearanceEnabled` -> `appearanceEnabled`

MREUnityRuntime/MREUnityRuntimeLib/Core/User.cs
stevenvergenz: It's the "default" group, that you're only in if you're not in any other groups. That relationship is maintained SDK-side, so no special logic in Unity.

---
In reply to: [267969008](https://github.com/Microsoft/mixed-reality-extension-unity/pull/60#discussion_r267969008) [](ancestors = 267969008)

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
stevenvergenz: Weird, I would have sworn I got just the opposite comment in a different PR, but I can't find it now. Changing it.

---
In reply to: [267969292](https://github.com/Microsoft/mixed-reality-extension-unity/pull/60#discussion_r267969292) [](ancestors = 267969292)

PR #61, opened 2019-03-22
Rename appearance patch field, convert signed ints to unsigned

stevenvergenz:
* Change patch field name to match Microsoft/mixed-reality-extension-sdk#272
* Add an unsigned int JSON converter, because JS doesn't know about that concept and sends down serialized signed ints

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Messaging/Payloads/Converters/UnsignedConverter.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Messaging/Payloads/Converters/UnsignedConverter.cs
@@ -16,7 +16,7 @@ namespace MixedRealityExtension.Messaging.Payloads.Converters
         /// <inheritdoc />
         public override bool CanConvert(Type objectType)
         {
-            return objectType == typeof(UInt32);
+            return UtilMethods.GetActualType(objectType) == typeof(UInt32);
         }
 
         /// <inheritdoc />
@@ -31,7 +31,8 @@ namespace MixedRealityExtension.Messaging.Payloads.Converters
             }
             else
             {
-                throw new JsonSerializationException($"Failed to deserialize {reader.Value}");
+                UnityEngine.Debug.Log($"Failed to deserialize {reader.ValueType} {reader.Value}");
+                return null;
             }
         }
 
--- a/MREUnityRuntime/MREUnityRuntimeLib/Messaging/Payloads/Converters/UnsignedConverter.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Messaging/Payloads/Converters/UnsignedConverter.cs
@@ -1,0 +1,47 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+using System;
+using System.Text;
+using MixedRealityExtension.Util;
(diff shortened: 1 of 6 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Messaging/Payloads/Converters/UnsignedConverter.cs
```diff
@@ -0,0 +1,47 @@
...
+using MixedRealityExtension.Util;
+using Newtonsoft.Json;
+
+namespace MixedRealityExtension.Messaging.Payloads.Converters
+{
+    /// <summary>
+    /// Converts signed integers into bit-equivalent unsigned integers,
+    /// because JS doesn't have a concept of unsigned.
+    /// </summary>
+    public class UnsignedConverter : JsonConverter
```
eanders-MS: Are you sure this works with nullable types? I had to account for nullable Enum types in the DashFormattedEnumConverter. I think you might need to do the same here.

tombuMS: You do need to account for nullable types.  The type check in CanConvert should use the same code that the DashFomrattedEnumConverter does.

MREUnityRuntime/MREUnityRuntimeLib/Patching/Types/UserPatch.cs
```diff
@@ -15,6 +17,7 @@ public class UserPatch : IPatchable
         public string Name { get; set; }
 
         [PatchProperty]
+        [JsonConverter(typeof(UnsignedConverter))]
```
eanders-MS: >UnsignedConverter [](start = 30, length = 17)

Is this actually necessary? Newtonsoft can't serialize unsigned ints natively?

MREUnityRuntime/MREUnityRuntimeLib/Patching/Types/AppearancePatch.cs
```diff
@@ -6,9 +6,13 @@ namespace MixedRealityExtension.Patching.Types
 {
     public class AppearancePatch
     {
-        [PatchProperty]
+        [JsonConverter(typeof(UnsignedConverter))]
         public UInt32? Enabled { get; set; }
```
eanders-MS: >Enabled [](start = 23, length = 7)

What's this field for?

stevenvergenz: I've commented it in a followup commit. It contains the same data as EnabledPacked, but it's sent on actor creation, and EnabledPacked is sent on update. This is because we don't have a mechanism to convert user-produced ActorLike objects into a different struct for transmission, and I prefer a clean API to clean transmission data.

MREUnityRuntime/MREUnityRuntimeLib/Patching/Types/UserPatch.cs
stevenvergenz: Newtonsoft can, but `JSON.parse` can't. Javascript doesn't have a concept of unsigned ints, so it stores the data as a signed int. As a result, what goes over the wire is "-1" instead of "4294967295".

PR #62, opened 2019-03-25
[Unity] User Masking take 2

stevenvergenz:
* Change `appearance.enabled` to `UInt32`
* Add `UInt32 IUser.Groups`
* Evaluate actor visibility against user group mask, re-evaluate when the user group mask changes
* Accept `user-update` messages from apps
* Remove deprecated `rigid-body-test` from testbed scene


MREUnityRuntime/MREUnityRuntimeLib/Core/ActorManager.cs
```diff
@@ -86,6 +86,14 @@ internal Actor FindActor(Guid id)
             }
         }
 
+        internal void UpdateAllVisibility()
+        {
+            foreach(var actor in _actorMapping.Values.Where(a => a.Parent == null))
+            {
+                Actor.ApplyVisibilityUpdate(actor, force: true);
+            }
```
eanders-MS: This needs to work on actor queues instead, queueing a local command to execute once the actor is loaded.

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -576,26 +590,41 @@ private void OnCreateFromPrefab(CreateFromPrefab payload, Action onCompleteCallb
         private void ProcessCreatedActors(CreateActor originalMessage, IList<Actor> createdActors, Action onCompleteCallback)
         {
             var guids = new DeterministicGuids(originalMessage.Actor?.Id);
+            var rootActor = createdActors.FirstOrDefault();
+
+            ProcessActors(rootActor.transform, rootActor.transform.parent.GetComponent<Actor>());
+
+            rootActor?.ApplyPatch(originalMessage.Actor);
```
eanders-MS: >rootActor? [](start = 12, length = 10)

Actor is a MonoBehaviour. Does that make this use of the `?` operator invalid here? (based on your comment in MixedRealityExtensionApp.cs)

PR #63, opened 2019-04-02
Transform API Update for App and Local Spaces

tombuMS:


MREUnityRuntime/MREUnityRuntimeLib/Core/Interfaces/IMixedRealityExtensionObject.cs
```diff
@@ -27,9 +27,9 @@ public interface IMixedRealityExtensionObject
         IUser LocalUser { get; }
```
eanders-MS: Unrelated, but: Every actor has a copy of the LocalUser? Seems weird.

eanders-MS: Also unrelated: I don't think we need the `IMixedRealityExtensionObject` interface. It's an unnecessary abstraction.

tombuMS: I think we should log an issue for the `LocalUser` thing as that seems like it could be an issue if we want to use this SDK in server authoritative as we discussed offline.  I will also remove the `IMixedRealityExtensionObject` interface.  I had started cleaning this up, but missed the interface.

tombuMS: As poer our discussion, I am leaving the `IMixedRealityExtensionObject` interface as it is used currently in the hots app code.  We can revisit later if the need arises.

PR #69, opened 2019-04-10
Fix for rigid body synchronization

tombuMS:
This is a fix for issue SDK:[#295](https://github.com/Microsoft/mixed-reality-extension-sdk/issues/295)


tombuMS:
This should also fix the issue with null reference exception in IsGrabbed on Occulus Go.  SDK:[#296](https://github.com/Microsoft/mixed-reality-extension-sdk/issues/296)

PR #71, opened 2019-04-11
Refactor usage of MixedRealityExtensionApp; fix build path error with spaces

influnza:
This change only refactors, no change to functionality. 
Removed IConnectionInternal that is not used.
Made several classes public: this will help integrating the part that handles the connection into another 3D engine potentially.
Replaced MixedRealityExtensionApp with its interface, where possible.

### DIFF
--- a/buildMREUnityRuntimeDLL.bat
+++ b/buildMREUnityRuntimeDLL.bat
@@ -1,7 +1,7 @@
 @rem Bootstrapping script to build MREUnityRuntimeLib dll
 @echo off
 
-for /f "usebackq tokens=*" %%i in (`%~dp0tools\vswhere -latest -products * -requires Microsoft.Component.MSBuild -property installationPath`) do (
+for /f "usebackq tokens=*" %%i in (`"%~dp0tools\vswhere" -latest -products * -requires Microsoft.Component.MSBuild -property installationPath`) do (
   set VSInstallDir=%%i
 )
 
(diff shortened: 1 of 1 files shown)


influnza:
> (As discussed IRL) Batch file changes are good, but it doesn't make sense to export additional symbols until such time we split the DLL into Unity and common implementations.

Reverted the change to visibility. Is the change to interfaces acceptable?

eanders-MS:
@tombuMS what are your thoughts? Is this a good change?

MREUnityRuntime/MREUnityRuntimeLib/App/IMixedRealityExtensionApp.cs
```diff
@@ -73,6 +74,11 @@ public interface IMixedRealityExtensionApp
         /// </summary>
         GameObject SceneRoot { get; set; }
 
+        /// <summary>
+        /// Current connection to the application server
+        /// </summary>
+        IConnection Connection { get; }
```
tombuMS: I don't think we should be exposing this to the host app.  The connection is supposed to be form internal use within the runtime.

MREUnityRuntime/MREUnityRuntimeLib/App/IMixedRealityExtensionApp.cs
```diff
@@ -133,5 +139,16 @@ public interface IMixedRealityExtensionApp
         /// </summary>
         /// <param name="actorId"></param>
         void OnActorDestroyed(Guid actorId);
+
+        /// <summary>
+        /// Send the event, using the current protocol
+        /// </summary>
+        /// <param name="update"></param>
+        void SendEvent(Payload update);
```
tombuMS: Sending events and working with payloads should not be exposed to host app implementations

MREUnityRuntime/MREUnityRuntimeLib/App/IMixedRealityExtensionApp.cs
```diff
@@ -133,5 +139,16 @@ public interface IMixedRealityExtensionApp
...
+
+        /// <summary>
+        /// Send the event, using the current protocol, with provide response ID
+        /// </summary>
+        void SendEvent(Payload update, string _responseId);
```
tombuMS: Same as above. This is an internal detail to the runtime that shouldn't be touched by host app integrations.

tombuMS:
> @tombuMS what are your thoughts? Is this a good change?

I don't think this is a change we want to take.  The purpose of the IMixedRealityExtensionApp interface is to expose only the subset of functionality that we want the host app to be able to override or needs to be able to call in to.  With that purpose in mind, we use the non-interface concrete class in our internal code so that we have access to the internal functionality that should remain only in the runtime, as it s not needed by the host app integrations.  We can consider exposing things in the interface on a case basis, but all up this change exposes much more than it should I believe.

influnza:
Removed the change to app interface.

eanders-MS:
Thank you @influnza.

PR #76, opened 2019-05-24
Ensure a final transform update from actor on end of grab

tombuMS:


tombuMS:
This is a fix for issue #[323](https://github.com/microsoft/mixed-reality-extension-sdk/issues/323)

PR #78, opened 2019-06-05
Tweak testbed for useful standalone builds

stevenvergenz:


MRETestBed/Assets/TestBed Assets/Scripts/UserInfo.cs
```diff
@@ -6,15 +6,17 @@
...
     {
         Id = id;
+        this.name = name;
     }
 
     public GameObject UserGO { get; set; }
 
     public Guid Id { get; }
 
+    private string name;
```
eanders-MS: Can you rename `name` so that it's not confused with the public `Name` accessor?

PR #81, opened 2019-06-17
Grab lerping on synchronized clients

tombuMS:
Unity work for SDK PR#(350)[https://github.com/microsoft/mixed-reality-extension-sdk/pull/350]

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
@@ -21,6 +21,7 @@ using UnityEngine;
 using UnityLight = UnityEngine.Light;
 using UnityCollider = UnityEngine.Collider;
 using MixedRealityExtension.PluginInterfaces.Behaviors;
+using MixedRealityExtension.Util;
 
 namespace MixedRealityExtension.Core
 {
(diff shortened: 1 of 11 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -774,55 +806,150 @@ private void PatchTransform(ActorTransformPatch transformPatch)
...
+                // Convert the app relative transform for the correction to world position relative to our app root.
+                Vector3? newPos = null;
+                Quaternion? newRot = null;
+
+                if (transform.Position != null)
+                {
+                    Vector3 appPos;
+                    appPos.x = transform.Position.X;
+                    appPos.y = transform.Position.Y;
+                    appPos.z = transform.Position.Z;
```
stevenvergenz: Are these vector patches partial? Do we need to check for nulls?

tombuMS: Actor corrections do not actually work off of patches.  Actor corrections will always contain a transform with full position and/or rotation to be lerped to.  Null checks are needed around the position and rotation are needed, but if they exist, all components in the vector 3 or quaternion will exist, by design.  Thanks for calling that out though to be sure.

PR #82, opened 2019-06-19
Fix issue where parent is deleted during a gltf async load.

tombuMS:
This fixes an issue where the GLTF being created has its parent destroyed in that process and we get in to a state where that gltf is still created, but never initialized and is now an orphan at the root of the scene.


tombuMS:
This is a fix for issue #[352](https://github.com/microsoft/mixed-reality-extension-sdk/issues/352)

PR #84, opened 2019-06-27
DO NOT MERGE: Video Player Factory interface abstraction

sorenhan:
As an alternative to the current AltspaceVR-exclusive RPC interface, this PR exposes a interface which Unity host apps can choose to implement, to enable streaming video players.

This interface is a work in progress - please comment, but don't merge in.

remaining work:
1. rename SetSoundState to SetMediaState (and possibly other SoundXXX names). - a bunch of sound commands are now shared with the video commands, but weren't renamed to match..
2. stop API isn't plumbed through from the SDK.
2. show/hide API not implemented

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Core/VideoStreamDescription.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Core/VideoStreamDescription.cs
@@ -1,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace MixedRealityExtension.Core
+{
+    public class VideoStreamDescription : UnityEngine.ScriptableObject
+    {
+        //Duration in seconds
+        public float Duration;
+    }
+}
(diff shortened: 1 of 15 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Core/VideoStreamDescription.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the MIT License.
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace MixedRealityExtension.Core
+{
+    public class VideoStreamDescription : UnityEngine.ScriptableObject
```
eanders-MS: >ScriptableObject [](start = 54, length = 16)

For my education: What is the benefit of deriving from ScriptableObject?

eanders-MS: Got my answer offline, thanks!

sorenhan: Documenting the answer: It is because our other asset manager is using UnityEngine.Object, not System.Object, and ScriptableObject is the right way to create a non-monobehavior object derived from UnityEngine.Object - UnityEngine.Object is a custom class that requires native engine functionality for instantiation

PR #87, opened 2019-07-12
Asset unloading

stevenvergenz:
* Add a `ContainerId` field to `load-assets` and `create-asset` messages.
* Add an `unload-assets` message and handler
* Rewrite asset cache to look up by any associated field.
* Fix bug where testbed resource factory doesn't throw correctly.


MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetLoader.cs
```diff
@@ -145,35 +146,14 @@ private async Task LoadAssets(LoadAssets payload, Action onCompleteCallback)
             string failureMessage = null;
 
             // attempt to get cached assets instead of loading
```
eanders-MS: >// attempt to get cached assets instead of loading [](start = 12, length = 50)

Delete stale comment.

PR #88, opened 2019-07-16
Fix patching during actor creation

stevenvergenz:
Fixes microsoft/mixed-reality-extension-sdk#386

* Get rid of actor update queues. There is already the network buffer and command queues in the actor manager. We don't need a third queue specifically for an actor's updates, it just makes the updates late.
* Process the actor command backlog before returning newly created actors in an `object-spawned` message. This prevents the returned patches from being out of date.

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Core/ActorManager.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Core/ActorManager.cs
@@ -104,13 +104,8 @@ namespace MixedRealityExtension.Core
 
         internal void ProcessActorCommand(Guid actorId, NetworkCommandPayload payload, Action onCompleteCallback)
         {
-            _actorCommandQueues.GetOrCreate(actorId, () =>
-            {
-                var queue = new ActorCommandQueue(actorId, _app);
-                _actorCommandQueues.Add(actorId, queue);
-                return queue;
-            })
-            .Enqueue(payload, onCompleteCallback);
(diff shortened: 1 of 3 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Core/ActorManager.cs
```diff
@@ -103,12 +104,21 @@ internal bool HasActor(Guid? id)
 
         internal void ProcessActorCommand(Guid actorId, NetworkCommandPayload payload, Action onCompleteCallback)
         {
-            if (!_actorCommandQueues.TryGetValue(actorId, out ActorCommandQueue queue))
+            _actorCommandQueues.GetOrCreate(actorId, () =>
             {
-                queue = new ActorCommandQueue(actorId, _app);
+                var queue = new ActorCommandQueue(actorId, _app);
                 _actorCommandQueues.Add(actorId, queue);
```
eanders-MS: >_actorCommandQueues.Add(actorId, queue); [](start = 16, length = 40)

Do you need to call Add here, or does GetOrCreate handle the addition?

stevenvergenz: GetOrCreate does everything for you, it's a util extension function that I copied over from Altspace ages ago.

---
In reply to: [304130289](https://github.com/microsoft/mixed-reality-extension-unity/pull/88#discussion_r304130289) [](ancestors = 304130289)

PR #90, opened 2019-07-22
Type conversion allocation refactor

tombuMS:
Removed unnecessary allocations associate with type conversions between Unity types and MRE types.

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
@@ -406,21 +406,23 @@ namespace MixedRealityExtension.Core
             };
         }
 
-        private MWScaledTransform localTransform = new MWScaledTransform();
-        private MWTransform appTransform = new MWTransform();
+        // These two variables are for local use in the SendActorUpdate method to prevent unnecessary allocations.  Their
+        // user should be limited to this function.
+        private MWScaledTransform __methodVar_localTransform = new MWScaledTransform();
+        private MWTransform __methodVar_appTransform = new MWTransform();
         internal void SendActorUpdate(ActorComponentType flags)
         {
             ActorPatch actorPatch = new ActorPatch(Id);
 
             if (flags.HasFlag(ActorComponentType.Transform))
             {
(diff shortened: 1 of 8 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -395,16 +406,21 @@ internal OperationResult EnableText(TextPatch textPatch)
             };
         }
 
+        private MWScaledTransform localTransform = new MWScaledTransform();
+        private MWTransform appTransform = new MWTransform();
```
eanders-MS: In this class, we have `AppTransform`, `_appTransform`, and `appTransform`. Can you differentiate them through renaming, or clarify their usage in comments?

tombuMS: Standard naming conventions is __name_ for private member variables, and no underscore for "local" variables which is what these are replacing.  I will add a comment and I can do a double or triple underscore.  That should really make it clear not to use them elsewhere in the class.

eanders-MS: A custom naming convention for temporary vars that stick around is good; brings a nice increase in clarity. Let's use this convention everywhere (not just in Actor.cs). Minor nit: `__methodVar_` seems excessive, maybe `_tmp_`?

---
In reply to: [306007763](https://github.com/microsoft/mixed-reality-extension-unity/pull/90#discussion_r306007763) [](ancestors = 306007763)

tombuMS: I think that `__methodVar_` is more descriptive than tmp.  I also thing the verbosity is good for quickly identifying the usage of the variable from intellisense.  If you feel strongly enough I could be convinced for `__tmp_`, though i still would insist on a double underscore than single.  It will filter them out as you type a member variable with a single underscore.

eanders-MS: `__methodVar_` it is.

PR #92, opened 2019-07-24
Add asset cache callback

stevenvergenz:
Add a mechanism to wait for an asset to exist. Fixes microsoft/mixed-reality-extension-sdk#393 and microsoft/mixed-reality-extension-sdk#394

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
@@ -579,7 +579,7 @@ namespace MixedRealityExtension.App
             {
                 MREAPI.AppsAPI.AssetCache.OnCached(payload.PrefabId, _ =>
                 {
-                    if (this == null) return;
+                    if (this == null || _appState != AppState.Running) return;
                     var createdActors = _assetLoader.CreateFromPrefab(payload.PrefabId, payload.Actor?.ParentId);
                     ProcessCreatedActors(payload, createdActors, onCompleteCallback);
                 });
--- a/MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
@@ -579,7 +579,7 @@ namespace MixedRealityExtension.App
             {
                 MREAPI.AppsAPI.AssetCache.OnCached(payload.PrefabId, _ =>
                 {
-                    if (!_ownerScript) return;
+                    if (this == null) return;
                     var createdActors = _assetLoader.CreateFromPrefab(payload.PrefabId, payload.Actor?.ParentId);
                     ProcessCreatedActors(payload, createdActors, onCompleteCallback);
                 });
--- a/MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
@@ -579,6 +579,7 @@ namespace MixedRealityExtension.App
             {
                 MREAPI.AppsAPI.AssetCache.OnCached(payload.PrefabId, _ =>
                 {
+                    if (!_ownerScript) return;
                     var createdActors = _assetLoader.CreateFromPrefab(payload.PrefabId, payload.Actor?.ParentId);
                     ProcessCreatedActors(payload, createdActors, onCompleteCallback);
                 });
--- a/MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
@@ -577,8 +577,11 @@ namespace MixedRealityExtension.App
         {
             try
             {
-                var createdActors = _assetLoader.CreateFromPrefab(payload.PrefabId, payload.Actor?.ParentId);
-                ProcessCreatedActors(payload, createdActors, onCompleteCallback);
+                MREAPI.AppsAPI.AssetCache.OnCached(payload.PrefabId, _ =>
+                {
+                    var createdActors = _assetLoader.CreateFromPrefab(payload.PrefabId, payload.Actor?.ParentId);
+                    ProcessCreatedActors(payload, createdActors, onCompleteCallback);
+                });
             }
             catch (Exception e)
             {
--- a/MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
@@ -53,6 +53,7 @@ namespace MixedRealityExtension.App
         }
 
         private AppState _appState = AppState.Stopped;
+        private int generation = 0;
 
         #region Events - Public
 
@@ -430,6 +431,7 @@ namespace MixedRealityExtension.App
 
         private void Conn_OnDisconnected()
         {
+            generation++;
             if (Protocol != null)
             {
                 Protocol.Stop();
@@ -577,11 +579,20 @@ namespace MixedRealityExtension.App
         {
             try
             {
-                MREAPI.AppsAPI.AssetCache.OnCached(payload.PrefabId, _ =>
+                var curGeneration = generation;
+                MREAPI.AppsAPI.AssetCache.OnCached(payload.PrefabId, prefab =>
                 {
-                    if (this == null || _appState != AppState.Running) return;
-                    var createdActors = _assetLoader.CreateFromPrefab(payload.PrefabId, payload.Actor?.ParentId);
-                    ProcessCreatedActors(payload, createdActors, onCompleteCallback);
+                    if (this == null || _appState != AppState.Running || generation != curGeneration) return;
+                    if (prefab != null)
+                    {
+                        var createdActors = _assetLoader.CreateFromPrefab(payload.PrefabId, payload.Actor?.ParentId);
+                        ProcessCreatedActors(payload, createdActors, onCompleteCallback);
+                    }
+                    else
+                    {
+                        var message = $"Prefab {payload.PrefabId} failed to load, cancelling actor creation";
+                        SendCreateActorResponse(payload, failureMessage: message, onCompleteCallback: onCompleteCallback);
+                    }
                 });
             }
             catch (Exception e)
--- a/MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
@@ -582,7 +582,7 @@ namespace MixedRealityExtension.App
                 var curGeneration = generation;
                 MREAPI.AppsAPI.AssetCache.OnCached(payload.PrefabId, prefab =>
                 {
-                    if (this == null || _appState != AppState.Running || generation != curGeneration) return;
+                    if (this == null || _conn == null || !_conn.IsActive || generation != curGeneration) return;
                     if (prefab != null)
                     {
                         var createdActors = _assetLoader.CreateFromPrefab(payload.PrefabId, payload.Actor?.ParentId);
--- a/MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetCache.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetCache.cs
@@ -8,6 +8,8 @@ using MixedRealityExtension.Util;
 using UnityEngine;
 using Object = UnityEngine.Object;
 
+using CacheCallback = System.Action<UnityEngine.Object>;
+
 namespace MixedRealityExtension.Assets
 {
     /// <summary>
@@ -32,6 +34,7 @@ namespace MixedRealityExtension.Assets
         }
 
         private readonly List<CacheEntry> cache = new List<CacheEntry>(50);
+        private readonly Dictionary<Guid, List<CacheCallback>> cacheCallbacks = new Dictionary<Guid, List<CacheCallback>>(50);
         private readonly GameObject cacheRoot;
         private readonly GameObject emptyTemplate;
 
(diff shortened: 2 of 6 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Factories/DefaultMaterialPatcher.cs
```diff
@@ -36,7 +36,7 @@ public void ApplyMaterialPatch(Material material, MWMaterial patch)
             }
             else if (patch.MainTextureId != null)
             {
-                material.mainTexture = (Texture)MREAPI.AppsAPI.AssetCache.GetAsset(patch.MainTextureId.Value);
+                MREAPI.AppsAPI.AssetCache.OnCached(patch.MainTextureId.Value, tex => material.mainTexture = (Texture)tex);
             }
```
sorenhan: This fixes defers setting the value, which is good, there is a risk that other changes to material.mainTexture may come in out of order. For example: If there is a call to set texture to an uncached texture instantly followed by setting texture to null (or a cached texture), then the uncached value would be set afterwards.

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -767,15 +767,10 @@ private void PatchAppearance(AppearancePatch appearance)
...
-                if (sharedMat != null)
+                MREAPI.AppsAPI.AssetCache.OnCached(MaterialId, sharedMat =>
                 {
-                    Renderer.sharedMaterial = sharedMat;
-                }
-                else
-                {
-                    MREAPI.Logger.LogWarning($"Material {MaterialId} not found, cannot assign to actor {Id}");
-                }
+                    Renderer.sharedMaterial = (Material)sharedMat;
```
sorenhan: >Renderer.sharedMaterial  [](start = 20, length = 24)

Same concern with order of deferred vs instant changes to a variable.

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetCache.cs
```diff
@@ -59,20 +62,41 @@ public GameObject EmptyTemplate()
...
+        {
+            var asset = GetAsset(id);
+            if (asset != null)
+            {
+                callback?.Invoke(asset);
+            }
+            else
+            {
+                cacheCallbacks.GetOrCreate(id, () => new List<CacheCallback>(3)).Add(callback);
+            }
```
sorenhan: I'd like if you could only call OnCached on assets that are actively being cached - currently you can use use OnCached on any unloaded asset ID, and it will just wait forever. Could we change CacheAsset to CacheAssetStart and CacheAssetComplete, and CacheAssetStart adds an entry so you're allowed to do OnCached on it, and otherwise it fails?

sorenhan: Similarly, if asset loading fails/times out, we should have a way to deal with this. CacheCallback may be called with null on failure?


---
In reply to: [306969021](https://github.com/Microsoft/mixed-reality-extension-unity/pull/92#discussion_r306969021) [](ancestors = 306969021)

stevenvergenz: Yes, waiting for an ID that will not be loaded would leak a little memory and cause the callback to never be executed. I don't think that can happen this far down the stack though, we have lots of checks in the SDK that the asset exists before we assign it. And unloaded assets don't exist for this purpose. If the user insists on doing something dumb, like generating a new guid and assigning that, it becomes `null` long before it hits Unity.

---
In reply to: [306969021](https://github.com/microsoft/mixed-reality-extension-unity/pull/92#discussion_r306969021) [](ancestors = 306969021)

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetCache.cs
```diff
@@ -59,20 +62,41 @@ public GameObject EmptyTemplate()
...
+        public void OnCached(Guid id, CacheCallback callback)
+        {
+            var asset = GetAsset(id);
+            if (asset != null)
+            {
+                callback?.Invoke(asset);
+            }
+            else
+            {
+                cacheCallbacks.GetOrCreate(id, () => new List<CacheCallback>(3)).Add(callback);
```
eanders-MS: `cacheCallbacks` needs to be cleared when app's `FreeResources` method is called.

MREUnityRuntime/MREUnityRuntimeLib/Factories/DefaultMaterialPatcher.cs
eanders-MS: Agreed, this implementation leads to buggy behavior. Changes should be applied in order.

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetCache.cs
```diff
@@ -59,20 +62,41 @@ public GameObject EmptyTemplate()
...
         /// <inheritdoc cref="CacheAsset"/>
         public void CacheAsset(Object asset, Guid id, Guid containerId, AssetSource source = null)
         {
             cache.Add(new CacheEntry(id, containerId, source, asset));
+            if (cacheCallbacks.TryGetValue(id, out List<CacheCallback> callbacks))
+            {
+                cacheCallbacks.Remove(id);
+                foreach (var cb in callbacks)
+                {
+                    cb?.Invoke(asset);
```
stevenvergenz: Wrap in try-catch

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetCache.cs
stevenvergenz: The cache doesn't know which callbacks are for which apps, so the behavior isn't this simple. Resolving the root concern by allowing the callbacks to run, but adding preconditions that return early if the app is destroyed.

---
In reply to: [307024446](https://github.com/microsoft/mixed-reality-extension-unity/pull/92#discussion_r307024446) [](ancestors = 307024446)

eanders-MS:
Please target 0.12.n branch. red branch has changes we don't want to take if we release this week.

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetCache.cs
```diff
@@ -59,20 +62,48 @@ public GameObject EmptyTemplate()
...
+            return asset != null ? cache.Find(c => c.Asset == asset).Id : (Guid?)null;
+        }
+
+        /// <inheritdoc cref="OnCached"/>
+        public void OnCached(Guid id, CacheCallback callback)
+        {
+            var asset = GetAsset(id);
+            if (asset != null)
+            {
+                callback?.Invoke(asset);
```
eanders-MS: Put a try/catch around this callback too.

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -577,8 +577,12 @@ private void OnCreateFromPrefab(CreateFromPrefab payload, Action onCompleteCallb
         {
             try
             {
-                var createdActors = _assetLoader.CreateFromPrefab(payload.PrefabId, payload.Actor?.ParentId);
-                ProcessCreatedActors(payload, createdActors, onCompleteCallback);
+                MREAPI.AppsAPI.AssetCache.OnCached(payload.PrefabId, _ =>
+                {
+                    if (!_ownerScript) return;
```
eanders-MS: _ownerScript doesn't get destroyed when the app disconnects. I think you'll need to do this another way. One idea: Keep a counter and increment it each time Conn_Disconnected is called. Check that the counter has the expected value in the callback.

MREUnityRuntime/MREUnityRuntimeLib/Factories/DefaultMaterialPatcher.cs
stevenvergenz: Resolved by tracking what texture is supposed to be assigned, and returning early if things aren't right.

---
In reply to: [306960603](https://github.com/microsoft/mixed-reality-extension-unity/pull/92#discussion_r306960603) [](ancestors = 306960603)

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
stevenvergenz: How about `this == null`?

---
In reply to: [307060178](https://github.com/microsoft/mixed-reality-extension-unity/pull/92#discussion_r307060178) [](ancestors = 307060178)

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
stevenvergenz: Added preconditions.

---
In reply to: [306960857](https://github.com/microsoft/mixed-reality-extension-unity/pull/92#discussion_r306960857) [](ancestors = 306960857)

stevenvergenz:
We're not shipping Tom's refactor in 0.12.3?

---
In reply to: [514832364](https://github.com/microsoft/mixed-reality-extension-unity/pull/92#issuecomment-514832364) [](ancestors = 514832364)

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
eanders-MS: That doesn't work either. This object doesn't get destroyed on a disconnect.

---
In reply to: [307061161](https://github.com/microsoft/mixed-reality-extension-unity/pull/92#discussion_r307061161) [](ancestors = 307061161,307060178)

eanders-MS:
Correct, not shipping it in 0.12.n.

---
In reply to: [514836478](https://github.com/microsoft/mixed-reality-extension-unity/pull/92#issuecomment-514836478) [](ancestors = 514836478,514832364)

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -75,7 +75,12 @@ public MWScaledTransform LocalTransform
         {
             get
             {
-                _localTransform = _localTransform ?? transform.ToLocalTransform();
+                if (_localTransform == null)
+                {
+                    _localTransform = new MWScaledTransform();
+                    _localTransform.ToLocalTransform(transform);
+                }
```
eanders-MS: Looks like Tom's changes snuck in. Can you rebase your changes to 0.12.n?

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
eanders-MS: I would be surprised if checking _appState == AppState.Running will work. Did you test it with a second peer joining and syncing assets? AppState isn't set to running until after synchronization completes, and assets load before that time.

---
In reply to: [307063204](https://github.com/microsoft/mixed-reality-extension-unity/pull/92#discussion_r307063204) [](ancestors = 307063204,307061161,307060178)

eanders-MS: Also, the app could have gone from running -> stopped -> running while the prefab was loading. You need to know for certain in the callback that the websocket didn't disconnect and reconnect.

---
In reply to: [307066354](https://github.com/microsoft/mixed-reality-extension-unity/pull/92#discussion_r307066354) [](ancestors = 307066354,307063204,307061161,307060178)

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
eanders-MS: Re-activating this thread. Tom's code still present.

---
In reply to: [307063794](https://github.com/microsoft/mixed-reality-extension-unity/pull/92#discussion_r307063794) [](ancestors = 307063794)

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -577,8 +577,12 @@ private void OnCreateFromPrefab(CreateFromPrefab payload, Action onCompleteCallb
         {
             try
             {
-                var createdActors = _assetLoader.CreateFromPrefab(payload.PrefabId, payload.Actor?.ParentId);
-                ProcessCreatedActors(payload, createdActors, onCompleteCallback);
+                MREAPI.AppsAPI.AssetCache.OnCached(payload.PrefabId, _ =>
+                {
+                    if (this == null || _appState != AppState.Running) return;
```
eanders-MS: >appState != AppState.Running [](start = 41, length = 28)

The CodeFlow comment thread for this line got undocked and is at the file of this file now.

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetCache.cs
stevenvergenz: Resolved by caching a `null` value when the asset fails to load. Permanently marks that guid as a dud.

---
In reply to: [306973398](https://github.com/microsoft/mixed-reality-extension-unity/pull/92#discussion_r306973398) [](ancestors = 306973398,306969021)

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -577,8 +579,21 @@ private void OnCreateFromPrefab(CreateFromPrefab payload, Action onCompleteCallb
         {
             try
             {
-                var createdActors = _assetLoader.CreateFromPrefab(payload.PrefabId, payload.Actor?.ParentId);
-                ProcessCreatedActors(payload, createdActors, onCompleteCallback);
+                var curGeneration = generation;
+                MREAPI.AppsAPI.AssetCache.OnCached(payload.PrefabId, prefab =>
+                {
+                    if (this == null || _appState != AppState.Running || generation != curGeneration) return;
```
eanders-MS: >_appState != AppState.Running [](start = 40, length = 29)

_appState != AppState.Running isn't the right check to do here. State synchronization happens before the app enters the running state, and assets are loaded during state synchronization.

eanders-MS: I think _conn.IsActive would be a safer alternative.

---
In reply to: [307544285](https://github.com/microsoft/mixed-reality-extension-unity/pull/92#discussion_r307544285) [](ancestors = 307544285)

stevenvergenz: Fixed

---
In reply to: [307544541](https://github.com/microsoft/mixed-reality-extension-unity/pull/92#discussion_r307544541) [](ancestors = 307544541,307544285)

PR #95, opened 2019-07-25
Fix null ref exception in rigid body update

tombuMS:
Fix for issue #[401](https://github.com/microsoft/mixed-reality-extension-sdk/issues/401).

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Core/RigidBody.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Core/RigidBody.cs
@@ -23,10 +23,10 @@ namespace MixedRealityExtension.Core
         private Queue<Action<Rigidbody>> _updateActions = new Queue<Action<Rigidbody>>();
 
         /// <inheritdoc />
-        public MWVector3 Velocity { get; set; }
+        public MWVector3 Velocity { get; set; } = new MWVector3();
 
         /// <inheritdoc />
-        public MWVector3 AngularVelocity { get; set; }
+        public MWVector3 AngularVelocity { get; set; } = new MWVector3();
 
         /// <inheritdoc />
         public float Mass { get; set; }


MREUnityRuntime/MREUnityRuntimeLib/Core/RigidBody.cs
```diff
@@ -23,10 +23,10 @@ internal class RigidBody : IRigidBody
         private Queue<Action<Rigidbody>> _updateActions = new Queue<Action<Rigidbody>>();
 
         /// <inheritdoc />
-        public MWVector3 Velocity { get; set; }
+        public MWVector3 Velocity { get; set; } = new MWVector3();
```
eanders-MS: Can `set;` be removed? Do you need to have the property setter?

tombuMS: Yes.  We set the velocity from the patching utils method.

eanders-MS: Is that in a different file? I only see Velocity.FromUnityVector3 in the Update method.

---
In reply to: [307493966](https://github.com/microsoft/mixed-reality-extension-unity/pull/95#discussion_r307493966) [](ancestors = 307493966)

eanders-MS: BTW, not a blocker, just was wondering. PR approved.

---
In reply to: [307547602](https://github.com/microsoft/mixed-reality-extension-unity/pull/95#discussion_r307547602) [](ancestors = 307547602,307493966)

tombuMS: Yes.  It is in the PatchingUtils.cs file for generating the rigid body patches.

PR #96, opened 2019-07-26
queue all mediainstance commands

SorenHannibal:
queue all mediainstance commands (start/update/stop) while asset is loading.

I like the idea of creating the managing class as soon as the command comes in, so I can see if the item is being loaded or is not valid. This is the same general structure I think the asset caching could benefit from, too - then we can discard/warn if we try to use assets that haven't even been preloaded, we can execute a callback if an asset load times out.


SorenHannibal:
(I'm not sure we actually want this in 0.12.n - we can just pull it into red, once the parent commit is also in red.

PR #97, opened 2019-07-27
per-app logging

SorenHannibal:
currently unity-side logging is printed to the unity log only. Ideally MRE Authors should be able to see the unity errors.  This PR adds a per-app logger class, which sends every logged message back to the MRE.

This is the Unity side of the PR. It works without SDK change.

We should also make the server say which level of logging it cares to receive - no need to waste bandwidth on diagnostics data that isn't looked at.

Caveat: This was only added to places that can access the App (maybe 65% of all places) - a number of static functions don't know which app they work for (or work for more than one app), so they will keep the old logger - most of those don't need to be brought over..

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Behaviors/BehaviorHandlerFactory.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Behaviors/BehaviorHandlerFactory.cs
@@ -34,8 +34,12 @@ namespace MixedRealityExtension.Behaviors
                     throw new NotImplementedException($"A handler for the behavior type {behaviorType.ToString()} exists but does not have a static create method implemented on it.");
                 }
             }
-            
-            actor.App.Logger.LogError($"Trying to create a behavior of type {behaviorType.ToString()}, but no handler is registered for the given type.");
+
+            MixedRealityExtensionApp app;
+            if (appRef.TryGetTarget(out app))
+            {
+                app.Logger.LogError($"Trying to create a behavior of type {behaviorType.ToString()}, but no handler is registered for the given type.");
+            }
             return null;
         }
     }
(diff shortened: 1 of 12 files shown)


SorenHannibal:
updated to use existing traces functionality instead of duplicating it.

MREUnityRuntime/MREUnityRuntimeLib/Behaviors/BehaviorHandlerFactory.cs
```diff
@@ -35,7 +35,7 @@ internal static IBehaviorHandler CreateBehaviorHandler(BehaviorType behaviorType
                 }
             }
             
-            MREAPI.Logger.LogError($"Trying to create a behavior of type {behaviorType.ToString()}, but no handler is registered for the given type.");
+            actor.App.Logger.LogError($"Trying to create a behavior of type {behaviorType.ToString()}, but no handler is registered for the given type.");
```
eanders-MS: >actor.App [](start = 12, length = 9)

I think you can just use `appRef` here rather than going through actor.

MREUnityRuntime/MREUnityRuntimeLib/Util/Logging/UnityLogger.cs
```diff
@@ -1,28 +1,46 @@
...
     internal class UnityLogger : IMRELogger
     {
+        MixedRealityExtensionApp _app;
+        public UnityLogger(MixedRealityExtensionApp app)
+        {
+            _app = app;
+        }
+        void Send(TraceSeverity severity, string message)
+        {
+            Traces traces = new Traces();
```
tombuMS: This is a lot of allocations here.  Every single log will now result in an allocation of the payload, the internal list of traces, and the single trace to send.  You should make a private member field that is the Traces payload and just reuse the one trace that will be in the list inside that payload.  Far more efficient.

eanders-MS:
>                                     MREAPI.Logger.LogError($"Trying to start sound instance that should already have completed for: {payload.MediaAssetId}\n");

I think you want to call `App.Logger` here, no?

---
Refers to: MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs:1265 in fd7a1fc. [](commit_id = fd7a1fc65e11aaad771bf2f60f335fe1981af275, deletion_comment = False)

PR #100, opened 2019-08-01
Don't log WebSocket.Send exceptions to App.Logger

eanders-MS:
Doing so will attempt to send a message over the WebSocket. If the WebSocket is disconnected, an infinite loop of errors can occur.


sorenhan:
Good call!

PR #101, opened 2019-08-08
Don't start prefab animations automatically

stevenvergenz:


eanders-MS:
Isn't this introducing a bug by changing the behavior configured in the model?

stevenvergenz:
Technically yes, but it's changing the behavior to match CreateFromGltf. I think the only reason this wasn't done from the beginning was we never tested it.

---
In reply to: [519613234](https://github.com/microsoft/mixed-reality-extension-unity/pull/101#issuecomment-519613234) [](ancestors = 519613234)

eanders-MS:
Did you consider the option to fix CreateFromGltf instead?

stevenvergenz:
CreateFromGltf does not require fixing, for a weird reason. Its animation component has `playAutomatically` set to true, but because of the way UnityGLTF creates the scene, the animation is empty when it gets created, so no animation is played. Because prefabs are fully created before they're instantiated, that flag causes us trouble.

---
In reply to: [519685399](https://github.com/microsoft/mixed-reality-extension-unity/pull/101#issuecomment-519685399) [](ancestors = 519685399)

eanders-MS:
What is the trouble exactly? Does the animation play on the prefab itself rather than on instances? Seems like we should respect this flag.

PR #105, opened 2019-08-22
Initial infrastructure for client side event handling

tombuMS:


eanders-MS:
Ok looks good so far!

MREUnityRuntime/MREUnityRuntimeLib/Behaviors/BehaviorHandlerBase.cs
```diff
@@ -45,6 +48,17 @@ protected void RegisterActionHandler(MWAction action, string name)
         {
             var handler = new BehaviorActionHandler(((IBehaviorHandler)this).BehaviorType, name, _appRef, _attachedActorId);
             action.Handler = handler;
+            _actionHandlers[name] = handler;
```
eanders-MS: >name [](start = 28, length = 4)

Do you need to lower-case this (and in GetActionHandler, below)? The value coming from the JSON payload will be lowercase.

eanders-MS: But the value from nameof(<enum value>) won't be, I believe.

---
In reply to: [316891965](https://github.com/microsoft/mixed-reality-extension-unity/pull/105#discussion_r316891965) [](ancestors = 316891965)

tombuMS: nameof does return it in the original case.  Good catch.

tombuMS:
Created a regular PR against feature/triggered-actions

PR #119, opened 2019-10-15
[Unity] Collision layers

stevenvergenz:
* Add `layer` field to ColliderPatches
* Add `collisionLayers` argument to InitializeAPI
* Add `collisionLayer` to CreateFromPrefab calls
* Fix `collider.enabled`. This property was misnamed `IsEnabled` in Unity, so would never be updated by the SDK, which calls it `enabled`.


MRETestBed/Assets/TestBed Assets/Scripts/MREComponent.cs
```diff
@@ -112,7 +112,8 @@ void Start()
 		if (!_apiInitialized)
 		{
 			MREAPI.InitializeAPI(
-				DefaultPrimMaterial,
+				defaultMaterial: DefaultPrimMaterial,
+				collisionLayers: new MixedRealityExtension.Core.EngineCollisionLayers(0, 9, 10, 5),
```
eanders-MS: >, 9, 10, 5 [](start = 75, length = 10)

Define constants for these. Also where did these come from?

eanders-MS: Oh wait this is the testbed, nevermind.

---
In reply to: [335213670](https://github.com/microsoft/mixed-reality-extension-unity/pull/119#discussion_r335213670) [](ancestors = 335213670)

MREUnityRuntime/MREUnityRuntimeLib/API/MREApi.cs
```diff
@@ -30,11 +31,13 @@ public static class MREAPI
...
 		/// <param name="gltfImporterFactory">The glTF loader factory. Uses default GLTFSceneImporter if omitted.</param>
 		/// <param name="materialPatcher">Overrides default material property map (color and mainTexture only).</param>
+		/// <param name="videoPlayerFactory"></param>
 		/// <param name="userInfoProvider">Provides appId/sessionId scoped IUserInfo instances.</param>
-		/// <param name="engineConstants">Engine constants supplied by the host app.</param>
+		/// <param name="dialogFactory"></param>
 		/// <param name="logger">The logger to be used by the MRE SDK.</param>
 		public static void InitializeAPI(
 			UnityEngine.Material defaultMaterial,
+			EngineCollisionLayers collisionLayers,
```
eanders-MS: >collisionLayers [](start = 25, length = 15)

Add docstring for collisionLayers.

MREUnityRuntime/MREUnityRuntimeLib/Messaging/Payloads/AssetCommandPayloads.cs
```diff
@@ -68,6 +68,8 @@ public class CreateFromPrefab : CreateActor
 		/// The ID of an already-loaded asset
 		/// </summary>
 		public Guid PrefabId;
+
+		public CollisionLayer? CollisionLayer;
```
eanders-MS: >CollisionLayer [](start = 9, length = 14)

Add docstring to CollisionLayer.

MREUnityRuntime/MREUnityRuntimeLib/Patching/Types/ColliderPatch.cs
```diff
@@ -10,13 +10,13 @@ namespace MixedRealityExtension.Patching.Types
 	public class ColliderPatch : IPatchable
```
eanders-MS: >ColliderPatch [](start = 14, length = 13)

This class should have docstrings, or make it internal.

stevenvergenz: All the patch classes are currently public and undocumented. Let's file a housekeeping bug.

---
In reply to: [335215069](https://github.com/microsoft/mixed-reality-extension-unity/pull/119#discussion_r335215069) [](ancestors = 335215069)

MREUnityRuntime/MREUnityRuntimeLib/Core/CollisionLayers.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace MixedRealityExtension.Core
```
tombuMS: The class and enum in this file are something that should be provided by the host app implementing MRE's.  All of these need docstrings for being exposed and used as public symbols in our runtime dll.

MREUnityRuntime/MREUnityRuntimeLib/UnityConstants.cs
```diff
@@ -4,6 +4,6 @@ namespace MixedRealityExtension
 {
 	internal static class UnityConstants
```
tombuMS: >internal static class UnityConstants [](start = 1, length = 36)

The main purpose for this file was for using it for layers.  Since we are taking a different approach to that, this file should be removed.

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetLoader.cs
```diff
@@ -78,7 +78,29 @@ internal IList<Actor> CreateFromPrefab(Guid prefabId, Guid? parentId)
 			var actorList = new List<Actor>();
 			MWGOTreeWalker.VisitTree(instance, go =>
 			{
-				go.layer = UnityConstants.ActorLayerIndex;
+				if (go.GetComponent<UnityEngine.Collider>() != null && collisionLayer != null)
```
tombuMS: >(go.GetComponent<UnityEngine.Collider>() != null [](start = 7, length = 48)

Not sure I completely understand why it has to have a collider to be assigned a layer, or even if I agree with this.  Why are we explicitly not allowing them to assign a layer to a game object when creating from a prefab at this point in the create process.  It seems unnecessarily restrictive as they could early assign a layer to any of the created prefabs and simply add a different collider type to each after the creation without having to know or assign the layer later on in the code.

Is this an attempt to remove the Unity like behavior that layers are associated with game objects rather than colliders?

stevenvergenz: Your last statement, pretty much. Unity uses layers for lots of different things, but we only use it for collisions. As such, it seemed to make sense to put the layer on the collider, like other engines do. There's no harm in putting collider-less actors on a layer, but it doesn't serve any purpose either. For consistency, I try to keep collider-less actors in the Default layer.

---
In reply to: [335525126](https://github.com/microsoft/mixed-reality-extension-unity/pull/119#discussion_r335525126) [](ancestors = 335525126)

tombuMS: I guess that is fine.

---
In reply to: [335597299](https://github.com/microsoft/mixed-reality-extension-unity/pull/119#discussion_r335597299) [](ancestors = 335597299,335525126)

PR #121, opened 2020-01-06
[Unity] Partial Animations API

stevenvergenz:
Tandem PR with microsoft/mixed-reality-extension-sdk#475.

* Update UnityGLTF to version that uses shortest-path quaternion slerping for animations.
* Add an `Animation` type and patch that defines an animation's properties.
* Add a `NativeAnimation` subclass of `Animation` that translates `Animation` onto `UnityEngine.AnimationState`s.
* Add an `AnimationManager` to track active instances and mutations.
* Send `AnimationPatch`es back in response to `create-from-prefab` and `create-animation` messages.
* Continue supporting old functionality from 0.14 apps and lower.


MREUnityRuntime/MREUnityRuntimeLib/Animation/AnimationManager.cs
```diff
@@ -0,0 +1,89 @@
...
+				// just write pending patch
+				PendingPatches[message.Animation.Id] = message.Animation;
+			}
+
+			onCompleteCallback?.Invoke();
+		}
+
+		public static long LocalUnixNow()
+		{
+			return (DateTime.UtcNow - Epoch).Ticks / 10_000;
```
eanders-MS: >10_000 [](start = 44, length = 6)

Cool. I learned something.

stevenvergenz: This syntax is available in typescript too, FYI.

---
In reply to: [368070517](https://github.com/microsoft/mixed-reality-extension-unity/pull/121#discussion_r368070517) [](ancestors = 368070517)

PR #122, opened 2020-02-22
Add ability for host apps to add GameObjects to an MRE

stevenvergenz:


MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -607,13 +659,30 @@ private void OnCreateFromPrefab(CreateFromPrefab payload, Action onCompleteCallb
...
 		{
-			var guids = new DeterministicGuids(originalMessage.Actor?.Id);
+			Guid guidGenSeed;
+			if (originalMessage != null)
+			{
+				guidGenSeed = originalMessage.Actor.Id;
+			}
+			else
+			{
+				var hasher = new System.Security.Cryptography.MD5CryptoServiceProvider();
```
eanders-MS: Use existing utility `UtilMethods.StringToGuid`

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -388,6 +388,35 @@ public void UpdateServerTimeOffset(long currentServerTime)
 			AnimationManager.UpdateServerTimeOffset(currentServerTime);
 		}
 
+		public void DeclarePreallocatedActors(GameObject[] objects, string guidSeed)
```
eanders-MS: We need to protect against this getting called multiple times, or allow it to be called multiple times safely, potentially with the same guidSeed.

eanders-MS: When is it not safe to execute this? We should protect against calling it too early by pending the operation.

eanders-MS: Will every client will call this when they load a space? If so, that could be a lot of redundant traffic. Perhaps delay the execution of this until after synchronization and skip actors already present at that time.

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -535,7 +535,7 @@ protected override void InternalUpdate()
 			}
 			catch (Exception e)
 			{
-				App.Logger.LogError($"Failed to synchronize app.  Exception: {e.Message}\nStackTrace: {e.StackTrace}");
+				App?.Logger.LogError($"Failed to synchronize app.  Exception: {e.Message}\nStackTrace: {e.StackTrace}");
```
eanders-MS: Why was this change necessary? This could should never execute if no App exists.

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
stevenvergenz: I like the idea of making sure that seeds aren't reused. Technically it's fine if it gets called multiple times, it's just less efficient because it sends multiple object-spawned messages.

stevenvergenz: Yes, each client will call this. We need to in order to register these gameobjects as actors locally, which we need for patching. You could make the case for not sending object-spawned, but that's not how the other actor creation flows work, and I'd like to keep it consistent.

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
stevenvergenz: This was to prevent log spam every frame if DeclarePreallocatedActors throws an exception.

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
stevenvergenz: I need to check. I called this during execution (onStarted) during testing, but maybe the sync phase would be better if it's supported.

eanders-MS: I have a concern over cross-platform support. This won't work for MREs that want to run across multiple platforms (Unreal & Altspace, for example). On the other hand, this does unlock a lot of capability for MREs authored just for Altspace. If we want to include this in the runtime API, we should be very clear about that.

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
eanders-MS: I don't see how App could ever be null. Now did you encounter that?

PR #123, opened 2020-02-26
Add in tool and pen behaviors in to runtime

tombuMS:
Unity runtime companion to SDK PR: #[484](https://github.com/microsoft/mixed-reality-extension-sdk/pull/484)

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Behaviors/Actions/Action.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Behaviors/Actions/Action.cs
@@ -71,6 +71,8 @@ namespace MixedRealityExtension.Behaviors.Actions
 
 		internal IActionHandler Handler { get; set; }
 
+		internal EventHandler<ActionStateChangedArgs> ActionStateChanging { get; set; }
+
 		internal EventHandler<ActionStateChangedArgs> ActionStateChanged { get; set; }
 
 		protected void ChangeState(IUser user, ActionState newState, BaseActionData actionData)
@@ -90,6 +92,7 @@ namespace MixedRealityExtension.Behaviors.Actions
 				_userActionStates.Add(user, newState);
 			}
 
+			ActionStateChanging?.Invoke(this, new ActionStateChangedArgs(user.Id, oldState, newState));
 			Handler?.HandleActionStateChanged(user, oldState, newState, actionData);
 			ActionStateChanged?.Invoke(this, new ActionStateChangedArgs(user.Id, oldState, newState));
 		}
--- a/MREUnityRuntime/MREUnityRuntimeLib/Behaviors/Actions/Action.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Behaviors/Actions/Action.cs
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+using MixedRealityExtension.Behaviors.ActionData;
 using MixedRealityExtension.Core.Interfaces;
 using System;
 using System.Collections.Generic;
@@ -19,13 +20,18 @@ namespace MixedRealityExtension.Behaviors.Actions
 		/// <summary>
 		/// The action is stopped.
 		/// </summary>
-		Stopped
+		Stopped,
+
+		/// <summary>
+		/// The action is currently being performed.
+		/// </summary>
+		Performing
 	}
 
 	/// <summary>
 	/// The event argument class to provide information about the action state change event.
 	/// </summary>
-	public sealed class ActionStateChangedArgs
+	internal sealed class ActionStateChangedArgs
 	{
 		/// <summary>
 		/// The id of the user that has cause a state change for the action.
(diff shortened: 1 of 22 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Behaviors/Handlers/PenHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+using MixedRealityExtension.App;
+using MixedRealityExtension.Behaviors.ActionData;
+using MixedRealityExtension.Behaviors.Actions;
+using MixedRealityExtension.Core.Interfaces;
+using MixedRealityExtension.PluginInterfaces.Behaviors;
+using System;
+
+namespace MixedRealityExtension.Behaviors.Handlers
+{
+	internal class PenToolHandler : ToolHandler<PenToolData>
```
eanders-MS: >PenToolHandler [](start = 16, length = 14)

Rename this file to match class name?

eanders-MS: Or, rename the class to PenHandler.

---
In reply to: [384833386](https://github.com/microsoft/mixed-reality-extension-unity/pull/123#discussion_r384833386) [](ancestors = 384833386)

MREUnityRuntime/MREUnityRuntimeLib/Behaviors/Handlers/PenHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+		internal PenToolHandler(IPenBehavior tool, WeakReference<MixedRealityExtensionApp> appRef, IActor attachedActor)
+			: base(tool, appRef, attachedActor)
+		{
+			
+		}
+
+		internal static PenToolHandler Create(IActor actor, WeakReference<MixedRealityExtensionApp> appRef)
+		{
+			var behaviorFactory = MREAPI.AppsAPI.BehaviorFactory;
+			var penBehavior = behaviorFactory.GetOrCreatePenBehavior(actor);
```
eanders-MS: >GetOrCreatePenBehavior [](start = 37, length = 22)

You're calling into application code here. It's good practice to protect against invalid return values and thrown exceptions.

MREUnityRuntime/MREUnityRuntimeLib/Behaviors/Actions/Action.cs
```diff
@@ -80,27 +126,53 @@ public void StartAction(IUser user)
...
 
-			Handler?.HandleActionStateChanged(user, oldState, newState);
+		/// <summary>
+		/// Signals the stop of this action for the given user.
+		/// </summary>
+		/// <param name="user">The user stopping the action.</param>
+		/// <param name="actionData">The optional data to pass along.</param>
+		public void StopAction(IUser user, ActionDataT actionData = null)
+		{
+			ChangeState(user, ActionState.Stopped, actionData);
```
eanders-MS: >ChangeState [](start = 3, length = 11)

I hope this question makes sense. When StopAction is called for the pen behavior, does it send the pen data accumulated since the last SynchronizeComponent call? Does it then clear out that data, or alternatively clear it on the next StartAction?

MREUnityRuntime/MREUnityRuntimeLib/Behaviors/Handlers/PenHandler.cs
tombuMS: Good catch.  I had renamed the file but not the class.

MREUnityRuntime/MREUnityRuntimeLib/Behaviors/Handlers/PenHandler.cs
tombuMS: Yeah.  We hadn't done that elsewhere in the behavior system, but can add that in now.

MREUnityRuntime/MREUnityRuntimeLib/Behaviors/Actions/Action.cs
tombuMS: Good catch.  Added in an assurance that the last perform action will be called before the stopping action is sent by the tool.

MREUnityRuntime/MREUnityRuntimeLib/Behaviors/Handlers/ToolHandler.cs
```diff
@@ -0,0 +1,68 @@
...
+			if (!IsUsing && wasUsing)
+			{
+				// We are stopping use and should send the remaining tool data up the remaining tool data from the last bit of use.
+				PerformUsingAction();
+			}
+		}
+
+		private void PerformUsingAction()
+		{
+			_queuedToolData = ToolData;
```
eanders-MS: >_queuedToolData [](start = 3, length = 15)

If _queuedToolData is empty, can we skip sending an update?

tombuMS: As a generic type we would need to introduce the concept of an empty too data.  I'll look in to adding this in quickly here or in a potential future PR related to this feature if a quick solution isn't found.

tombuMS: Added in this optimization.  Pretty easy to do in a general way.

PR #125, opened 2020-03-04
[Unity] Add an emissive color to materials

stevenvergenz:


MRETestBed/Assets/TestBed Assets/Scripts/VertexMaterialPatcher.cs
```diff
@@ -1,16 +1,69 @@
...
 		base.ApplyMaterialPatch(material, patch);
 
+		if (patch.EmissiveColor != null)
+		{
+			var color = material.GetColor(EmissiveColorProp);
+			color.r = patch.EmissiveColor.R ?? color.r;
+			color.g = patch.EmissiveColor.G ?? color.g;
+			color.b = patch.EmissiveColor.B ?? color.b;
+			color.a = patch.EmissiveColor.A ?? color.a;
+			material.SetColor(EmissiveColorProp, color);
```
stevenvergenz: Note: the patch utils are internal to the DLL, so they can't be used in factories.

PR #129, opened 2020-03-28
Update buildMREUnityRuntimeDLL.bat

radWorx:
Updated for VS 2019 and successfully ran Hello world in VS 2019.

### DIFF
--- a/buildMREUnityRuntimeDLL.bat
+++ b/buildMREUnityRuntimeDLL.bat
@@ -5,7 +5,7 @@ for /f "usebackq tokens=*" %%i in (`"%~dp0tools\vswhere" -products * -requires M
   set VSInstallDir=%%i
   if exist "%%i\MSBuild\Current\Bin\MSBuild.exe" goto FoundVisualStudio
 )
-echo MSBuild not found - please install Visual Studio 2019
+echo MSBuild not found - please install Visual Studio 2017 or later
 pause
 exit /b 1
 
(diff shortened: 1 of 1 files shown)


buildMREUnityRuntimeDLL.bat
```diff
@@ -3,20 +3,20 @@
 
 for /f "usebackq tokens=*" %%i in (`"%~dp0tools\vswhere" -products * -requires Microsoft.Component.MSBuild -property installationPath`)	 do (
   set VSInstallDir=%%i
-  if exist "%%i\MSBuild\15.0\Bin\MSBuild.exe" goto FoundVisualStudio
+  if exist "%%i\MSBuild\Current\Bin\MSBuild.exe" goto FoundVisualStudio
```
eanders-MS: This is a good change.

buildMREUnityRuntimeDLL.bat
```diff
@@ -3,20 +3,20 @@
...
 )
-echo MSBuild not found - please install Visual Studio 2017
+echo MSBuild not found - please install Visual Studio 2019
```
eanders-MS: I'd prefer this be changed to `2017 or later`, because Unity itself is officially still on 2017. Same comment applies below.

radWorx: Got it. thank you.

eanders-MS:
Thanks for taking the time to submit this!

PR #133, opened 2020-04-01
[WIP] Relative animations

stevenvergenz:
* Destroy animations when the last of its targets is destroyed, or the animation data is unloaded.
* Add ability for actors to generate partial patches, and reuse patch instances.
* Capture the state of targeted actor fields when anims get turned on.
* If an anim track is marked as `relative`, add interpolated values to the saved base values.
* If an anim's first keyframe time is greater than 0, use the saved values as an implicit first keyframe at t=0.


MREUnityRuntime/MREUnityRuntimeLib/Animation/Animation.cs
```diff
@@ -91,88 +120,49 @@ internal override void Update()
...
+				Interpolations.Interpolate(prevFrame.Value, nextFrame.Value, linearT, ref interpolatedToken, nextFrame.Bezier ?? track.Bezier);
+
+				// mix starting values with relative keyframe values
+				JToken mixedToken = interpolatedToken;
+				if (track.Relative == true)
 				{
-					jToken = jValue;
-					Interpolations.Interpolate(prevFrame.Value, nextFrame.Value, linearT, ref jToken, nextFrame.Bezier ?? track.Bezier);
+					mixedToken = TokenPool.Lease(interpolatedToken.Type);
+					Interpolations.ResolveRelativeValue(ImplicitStartKeyframes[ti].Value, interpolatedToken, ref mixedToken);
```
eanders-MS, line 146: I think you can preprocess the animation so that this doesn't need to be called every frame.

MREUnityRuntime/MREUnityRuntimeLib/Patching/Types/TransformPatch.cs
```diff
@@ -136,7 +161,7 @@ internal ScaledTransformPatch(MWVector3 position, MWQuaternion rotation, MWVecto
 			Scale = new Vector3Patch(scale);
 		}
 
-		void IPatchable.WriteToPath(TargetPath path, JToken value, int depth)
+		public void WriteToPath(TargetPath path, JToken value, int depth)
```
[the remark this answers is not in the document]
stevenvergenz: You're right. This is fixed in a subsequent PR.

MREUnityRuntime/MREUnityRuntimeLib/Animation/Interpolations.cs
```diff
@@ -73,5 +72,80 @@ internal static void Interpolate(JToken a, JToken b, float linearT, ref JToken m
 				}
 			}
 		}
+
+		internal static void ResolveRelativeValue(JToken reference, JToken relative, ref JToken result)
+		{
+			// compound types
+			if (reference.Type == JTokenType.Object)
+			{
+				JObject Reference = (JObject)reference;
```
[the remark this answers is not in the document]
stevenvergenz: The scale test (one 20-track animation) generates ~55KB of garbage per frame in the Test Bed, and the dynamic test (one 1-track animation) generates ~33KB. Compare that to ~25KB for the grab test. So animations generate a non-trivial amount of garbage, but it's not outrageous. This seems more like an optimization than a blocking issue to me.

MREUnityRuntime/MREUnityRuntimeLib/Animation/Animation.cs
stevenvergenz: That's true, we could produce absolute keyframe values from the relative ones on animation start, then interpolate between those, instead of resolving relative values every frame. Good idea!

eanders-MS: I'm not sure taking a snapshot of the starting value is a good idea. I wouldn't expect that from this API design, I'd like to support things like implementing a follow behavior as a looping animation. Let's discuss.

MREUnityRuntime/MREUnityRuntimeLib/Animation/Interpolations.cs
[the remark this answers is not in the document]
eanders-MS: Fair point. We can optimize it later.

MREUnityRuntime/MREUnityRuntimeLib/Animation/Animation.cs
eanders-MS: Let's leave it for now. We can improve it in an optimization pass.

eanders-MS: Just to expand on my comment, if this method instead returned a closure that returns the desired field, it could be cached in the keyframe. That way you'd avoid having to re-resolve the referenced field each frame.

stevenvergenz: Are we talking about the same thing? I don't think you could implement a follow behavior with this. This is where you add the interpolated keyframe value to the animating object's original property value, as it was at the start of the animation.

PR #135, opened 2020-04-14
[Unity] Animations overhaul part 2

stevenvergenz:
* Add new asset type `AnimationDataCached`, created by `CreateAnimation2` messages, plus all contained types: `Track` and `Keyframe`.
* Add new implementation of the `BaseAnimation` class (formerly just `Animation`) called `Animation`, which computes patches to apply to actors every frame.
* Overhaul patch types to support rapid reuse without generating excessive garbage memory.
* Update `Actor.GeneratePatch` to support partial patches.

Contains previously reviewed code from #133 and #134.


MREUnityRuntime/MREUnityRuntimeLib/Animation/JTokenPool.cs
```diff
@@ -0,0 +1,91 @@
...
+					return new JObject()
+					{
+						{"x", new JValue(0) },
+						{"y", new JValue(0) },
+						{"z", new JValue(0) }
+					};
+				case TokenPoolType.Value:
+					return new JValue(0);
+				default:
+					throw new System.Exception("How did we get here? Invalid enum value.");
```
eanders-MS, line 87: Throw a `NotSupportedException` here.

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -339,58 +339,128 @@ internal void Destroy()
...
+				output.RestoreAll();
+			}
 
-			ColliderPatch collider = null;
-			_collider = gameObject.GetComponent<UnityCollider>();
-			if (_collider != null)
+			if (generateAll || path.PathParts[0] == "transform")
 			{
-				if (Collider == null)
+				if (generateAll || path.PathParts[1] == "local")
```
eanders-MS, line 362: Embedding the target path schema in code like this will create a maintenance headache. We need to take a data-driven approach. You have the seed of it in TargetPath.cs: The TokenLookup class maps a value to a path. You could convert that  value to a utility class tailored to working on the specific type and value of the mapped target path.

MREUnityRuntime/MREUnityRuntimeLib/Animation/Interpolations.cs
```diff
@@ -0,0 +1,180 @@
...
+			var easedT = easing.Sample(linearT);
+
+			// compound types
+			if (a.Type == JTokenType.Object)
+			{
+				JObject A = (JObject)a;
+				JObject B = (JObject)b;
+				JObject Mix = (JObject)mix;
+
+				if (A.ContainsKey("x") && A.ContainsKey("y"))
```
eanders-MS, line 27: The interpolator should not have to inspect the value and rediscover how to interpolate its value every frame. I recommend making a set of interpolator functions, one for each type: QuaternionInterpolator, Vector3Interpolator, ValueInterpolator, etc. It should be possible using targetpath to know ahead of time which interpolator to use, then cache that interpolator method somewhere appropriate, perhaps in the keyframe or on the track.

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
stevenvergenz: I don't follow. The reason this method is so bad is because the internal structure of Actor is sufficiently different from ActorPatch that there's no way to map them except explicitly, which is what I've done here.

eanders-MS: I want to find a solution to this that is schema-driven, so that when the schema changes, this code won't compile until it is fixed up to match the updated schema. As it is, it will be a source of bugs over time. But we can ship this as-is for now. There's a bigger clean up needed this can be a part of.

PR #136, opened 2020-04-16
Add physics bridge

nik0la31:


MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -155,6 +157,7 @@ internal MixedRealityExtensionApp(string globalAppId, MonoBehaviour ownerScript,
 			_assetLoader = new AssetLoader(ownerScript, this);
 			_userManager = new UserManager(this);
 			_actorManager = new ActorManager(this);
+			_physicsBridge = new PhysicsBridge();
```
jabenk, line 160: Might we need here "this" for the physics bridge ?

MREUnityRuntime/MREUnityRuntimeLib/Core/PhysicsBridge.cs
```diff
@@ -0,0 +1,229 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace MixedRealityExtension.Core
+{
+	public class PhysicsBridge
```
jabenk, line 9: I think some code comments would be usefull

jabenk: on the functions level and maybe not for all functions

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
nik0la31: There was not need for it yet. We may add it later.

MREUnityRuntime/MREUnityRuntimeLib/Core/PhysicsBridge.cs
nik0la31: Agree, that's a big todo.

nik0la31:
Added initial physics bridge classes.

PR #137, opened 2020-04-25
Don't use conditional accessor on MonoBehaviors

eanders-MS:


MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -66,9 +66,31 @@ public MediaInstance(Guid MediaAssetId)
...
-		private MeshFilter filter = null;
-		internal MeshFilter MeshFilter => filter = filter ?? GetComponent<MeshFilter>();
+		internal Renderer Renderer
+		{
+			get
+			{
+				if (renderer == null)
+				{
+					var t = GetComponent<Renderer>();
+					renderer = t;
```
stevenvergenz, line 76: Why have the temp variable? Here and below.

eanders-MS: Oh that was for debugging. Will make a note to remove.

PR #141, opened 2020-04-28
Jabenk/physics merge

jabenk:
- Red merge + bounciness and friction


jabenk:
As agreed with Nikola I will PR this. All the tests seem to work.

PR #146, opened 2020-05-02
Limit concurrent web requests

eanders-MS:
This is a quick fix to stabilize AltQuiz. Globally limit the number of concurrent asset fetch web requests in the MRE runtime.

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetLoader.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetLoader.cs
@@ -146,36 +146,21 @@ namespace MixedRealityExtension.Assets
 			onCompleteCallback?.Invoke();
 		}
 
-		// Global: Keep track of the number of active glTF web requests.
-		private static int activeLoads = 0;
-
 		private async Task<IList<Asset>> LoadAssetsFromGLTF(AssetSource source, Guid containerId, ColliderType colliderType)
 		{
-			// Spin asynchronously until activeLoads allows us through.
-			while (activeLoads >= 4)
-			{
-				await Task.Delay(10);
-			}
-
 			WebRequestLoader loader = null;
 			Stream stream = null;
 			IList<Asset> assets = new List<Asset>();
 			DeterministicGuids guidGenerator = new DeterministicGuids(UtilMethods.StringToGuid(
 				$"{containerId}:{source.ParsedUri.AbsoluteUri}"));
-			try
-			{
-				// Increment activeLoads
-				++activeLoads;
 
+			// Wait asynchronously until the load throttler lets us through.
+			using (var scope = await AssetLoadThrottling.AcquireLoadScope())
+			{
 				// download file
 				var rootUrl = URIHelper.GetDirectoryName(source.ParsedUri.AbsoluteUri);
 				loader = new WebRequestLoader(rootUrl);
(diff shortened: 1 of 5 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Util/Unity/AssetFetcher.cs
```diff
@@ -67,13 +81,16 @@ IEnumerator LoadCoroutine()
...
 						}
 						else if (typeof(T) == typeof(Texture))
 						{
-							result.Asset = ((DownloadHandlerTexture) handler).texture as T;
+							result.Asset = ((DownloadHandlerTexture)handler).texture as T;
 						}
 					}
+
+					// Decrement activeLoads
+					Interlocked.Decrement(ref activeLoads);
```
eanders-MS, line 93: Would be nice to be able to decrement the counter before the finalization handler was called.

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetLoader.cs
```diff
@@ -146,125 +146,153 @@ private async Task LoadAssets(LoadAssets payload, Action onCompleteCallback)
 			onCompleteCallback?.Invoke();
 		}
 
+		// Global: Keep track of the number of active glTF web requests.
+		private static int activeLoads = 0;
+
 		private async Task<IList<Asset>> LoadAssetsFromGLTF(AssetSource source, Guid containerId, ColliderType colliderType)
 		{
+			// Spin asynchronously until activeLoads allows us through.
+			while (activeLoads >= 4)
```
StLoch, line 155: wbn to have the '4' as a shared constant for easy tuning / readability.

eanders-MS: Good call. Updated.

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetLoader.cs
```diff
@@ -148,123 +148,135 @@ private async Task LoadAssets(LoadAssets payload, Action onCompleteCallback)
...
 
-			GLTFSceneImporter importer =
-				MREAPI.AppsAPI.GLTFImporterFactory.CreateImporter(gltfRoot, loader, _asyncHelper, stream);
-			importer.SceneParent = MREAPI.AppsAPI.AssetCache.CacheRootGO().transform;
-			importer.Collider = colliderType.ToGLTFColliderType();
-
-			// load textures
-			if (gltfRoot.Textures != null)
+			using (GLTFSceneImporter importer =
+				MREAPI.AppsAPI.GLTFImporterFactory.CreateImporter(gltfRoot, loader, _asyncHelper, stream))
```
stevenvergenz, line 187: Oh good catch, I didn't realize glTF loaders were disposable. Are we sure disposing this doesn't destroy the in-memory assets that are in-use?

eanders-MS: I haven't changed the lifetime of the loader. It was already being disposed, but the call was explicit (line 266 on the left-hand side). This is just a refinement to make the Dispose call implicit, but also makes the lifetime tolerant against exceptions.

PR #147, opened 2020-05-04
Brwils/performance investigations

stevenvergenz:


MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetCache.cs
```diff
@@ -38,6 +38,7 @@ public CacheEntry(Guid id, Guid containerId, AssetSource source, Object asset, C
 
 		private readonly List<CacheEntry> cache = new List<CacheEntry>(50);
 		private readonly Dictionary<Guid, List<CacheCallback>> cacheCallbacks = new Dictionary<Guid, List<CacheCallback>>(50);
+		private readonly Dictionary<Guid, Tuple<GLTFSceneImporter, GLTF.Schema.GLTFRoot>> importerCache = new Dictionary<Guid, Tuple<GLTFSceneImporter, GLTF.Schema.GLTFRoot>>(50);
```
stevenvergenz, line 41: There's no point holding onto the importers after we've pulled all the assets out. Better to dispose of these immediately and free up the memory, then if we get a request to load this asset again, simply return the assets generated last time. So maybe holding onto URL to Guid[] would be better.

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetLoader.cs
```diff
@@ -146,35 +147,60 @@ private async Task LoadAssets(LoadAssets payload, Action onCompleteCallback)
...
 			{
 				for (var i = 0; i < gltfRoot.Textures.Count; i++)
 				{
-					await importer.LoadTextureAsync(gltfRoot.Textures[i], i, true);
-					var texture = importer.GetTexture(i);
-					texture.name = gltfRoot.Textures[i].Name ?? $"texture:{i}";
+					Guid id = guidGenerator.Next();
+
+					UnityEngine.Texture texture = MREAPI.AppsAPI.AssetCache.GetAsset(id) as UnityEngine.Texture;
+					if (texture == null)
```
stevenvergenz, line 196: Ideally we'd never get this far. If assets already exist with the given ID, we should use those instead.

MREUnityRuntime/MREUnityRuntimeLib/IPC/IConnection.cs
```diff
@@ -87,6 +88,8 @@ public interface IConnection : IDisposable
 		/// </summary>
 		/// <param name="message"></param>
 		void Send(string message);
```
stevenvergenz, line 90: If IConnections are not operating on strings anymore, probably better to remove this overload.

MREUnityRuntime/MREUnityRuntimeLib/Patching/Patching.cs
```diff
@@ -1,32 +1,27 @@
...
-					else if (val != null)
+					if (IsPatched(val as IPatchable))
 					{
 						return true;
 					}
 				}
+				else if (val != null)
+				{
+					return true;
+				}
```
stevenvergenz, line 24: Shouldn't we combine all these "return true" conditions?

MREUnityRuntime/MREUnityRuntimeLib/Patching/Types/TransformPatch.cs
```diff
@@ -143,46 +143,104 @@ public virtual void RestoreAll()
 		}
 	}
 
-	public class ScaledTransformPatch : TransformPatch, IPatchable
+	/// <note>
+	/// Currently duplicates TransformPatch instead of inheriting so we can directly inherit from
+	/// Patchable to get those wins. Sadly C# does not support multiple inheritance or templated
+	/// mixins.
+	/// </note>
+	public class ScaledTransformPatch : Patchable<ScaledTransformPatch>, IPatchable
```
stevenvergenz, line 151: If there's no shared code anymore, let's move `ScaledTransformPatch` to its own file.

MREUnityRuntime/MREUnityRuntimeLib/IPC/Connections/WebSocket.cs
```diff
@@ -101,6 +103,55 @@ private async Task CloseInternal(bool clean, string reason)
...
+				}
+				catch (Exception e)
+				{
+					MREAPI.Logger.LogDebug($"Error serializing message. Exception: {e.Message}\nStackTrace: {e.StackTrace}");
+				}
+			});
+		}
+
 		/// <inheritdoc />
 		public void Send(string message)
```
eanders-MS, line 156: >Send [](start = 14, length = 4)

WebSocket changes look fine. I'd like to reimplement them in a separate branch and make some collateral changes, like removing `Send(string)`. Also, I'd like to test this change in isolation.

MREUnityRuntime/MREUnityRuntimeLib/Messaging/Commands/CommandHandlerContext.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Reflection;
+
+namespace MixedRealityExtension.Messaging.Commands
+{
+	/// <summary>
+	/// This class pre-caches method info for command handlers, since it does not change at runtime.
+	/// This saved us 20ms on the startup frame for the MREApp on a high end CPU.
+	/// </summary>
+	/// <typeparam name="T"></typeparam>
+	internal class CommandHandlerContext<T> : ICommandHandlerContext
```
eanders-MS, line 13: >CommandHandlerContext [](start = 16, length = 21)

This is interesting. @Tombu what do you think? We were discussing the removal of exactly this reflection earlier today. Look like a good solution?

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -815,8 +830,8 @@ void ProcessActors2(Transform xfrm)
 						Message = trace.Message
 					},
 					Traces = new List<Trace>() { trace },
-					Actors = actors?.Select((actor) => actor.GeneratePatch()) ?? new ActorPatch[] { },
-					Animations = anims?.Select(anim => anim.GeneratePatch()) ?? new AnimationPatch[] { }
+					Actors = actors?.Select((actor) => actor.GeneratePatch()).ToArray() ?? new ActorPatch[] { },
+					Animations = anims?.Select(anim => anim.GeneratePatch()).ToArray() ?? new AnimationPatch[] { }
```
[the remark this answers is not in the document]
eanders-MS: From the comments in Payload.cs, it seems that changing these to arrays was necessary due to the change in WebSocket to serialize payloads off-thread. On the flip side, I'll need to play with this change in isolation to understand how it can deserialize off-thread and not touch Unity objects.

---
In reply to: [419657447](https://github.com/microsoft/mixed-reality-extension-unity/pull/147#discussion_r419657447) [](ancestors = 419657447)

MREUnityRuntime/MREUnityRuntimeLib/Messaging/Commands/CommandHandlerContext.cs
tombuMS: It's still doing reflection here. just during the static constructor of the first instance of the object.  I still think that there is room to improve this further by removing the need for reflection altogether.

---
In reply to: [419882530](https://github.com/microsoft/mixed-reality-extension-unity/pull/147#discussion_r419882530) [](ancestors = 419882530)

MREUnityRuntime/MREUnityRuntimeLib/Messaging/Protocols/Protocol.cs
```diff
@@ -53,16 +50,10 @@ public void Stop()
...
 		}
 
-		private void Conn_OnReceive(string json)
+		private void Conn_OnReceive(Message message)
 		{
 			try
 			{
-#if ANDROID_DEBUG
-				App.Logger.LogDebug($"Recv: {json}");
-#endif
```
tombuMS, line 62: This is very useful logging for the android build and shouldn't just be removed.  Can this be added back in to print out the message now that this has changed to a message object?

MREUnityRuntime/MREUnityRuntimeLib/Messaging/Commands/CommandHandlerContext.cs
eanders-MS: Even so, it's now a one-time startup cost. That's a significant improvement.

---
In reply to: [421122983](https://github.com/microsoft/mixed-reality-extension-unity/pull/147#discussion_r421122983) [](ancestors = 421122983,419882530)

PR #149, opened 2020-05-05
Unload an app's asset containers when the app closes

stevenvergenz:


MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -259,6 +259,12 @@ private void FreeResources()
 			_ownedGameObjects.Clear();
 			_actorManager.Reset();
 			AnimationManager.Reset();
+
+			foreach (Guid id in _assetLoader.ActiveContainers)
+			{
+				MREAPI.AppsAPI.AssetCache.UncacheAssets(id);
```
eanders-MS, line 265: I tried this out in Altspace, but it didn't work. I set a breakpoint here (line 265), and when I stopped my MRE (chess), the breakpoint hit, but the instantiated assets under the "MRE Cache" GameObject weren't destroyed. When I re-enabled my MRE, they were loaded again, doubling the cache size.

stevenvergenz: Huh, you're right. The asset cache design returned the removed assets when they were un-cached, for the caller to destroy. I think this was done to clean up references or something, but that's all done SDK-side now. I've moved the destroy logic to a new method on the asset cache, and call it from both the unload handler and the onDisconnect handler.

eanders-MS: I tried it in Altspace again, this time with two instances of chess with the same sessionId. Stopping one of the instances resulted in the other instance losing all its visuals, but was still running and connected to the MRE. It seems multiple app instances can share a cache or container. Only half of the game objects in "MRE Cache" were destroyed (as expected).

---
In reply to: [420389883](https://github.com/microsoft/mixed-reality-extension-unity/pull/149#discussion_r420389883) [](ancestors = 420389883)

PR #151, opened 2020-05-06
Move json serialization off main thread

eanders-MS:
Import json serialization improvements from @StLoch as an isolated feature.

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Messaging/Protocols/Protocol.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Messaging/Protocols/Protocol.cs
@@ -53,16 +53,10 @@ namespace MixedRealityExtension.Messaging.Protocols
 			}
 		}
 
-		private void Conn_OnReceive(string json)
+		private void Conn_OnReceive(Message message)
 		{
 			try
 			{
-#if ANDROID_DEBUG
-				App.Logger.LogDebug($"Recv: {json}");
-#endif
-
-				var message = JsonConvert.DeserializeObject<Message>(json, Constants.SerializerSettings);
-
 				if (message.Payload is Payloads.Heartbeat payload)
 				{
 					App.UpdateServerTimeOffset(payload.ServerTime);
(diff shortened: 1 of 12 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Messaging/Protocols/Protocol.cs
```diff
@@ -53,16 +53,10 @@ public void Stop()
...
 
-		private void Conn_OnReceive(string json)
+		private void Conn_OnReceive(Message message)
 		{
 			try
 			{
-#if ANDROID_DEBUG
-				App.Logger.LogDebug($"Recv: {json}");
-#endif
-
```
tombuMS, line 63: This is useful debug logging for when android breaks and shouldn't be just removed I don't think.

tombuMS:
Does this work actually move the json serialization off the main thread?  Or is it just a subset of that, as I don't see anything moved off of the main thread.

eanders-MS:
It happens by virtue of the WebSocket implementation. Serialization happens on WebSocket's 'sendWorker' task. Similarly for deserialization, which happens in WebSocket's 'workerTask' task.

---
In reply to: [624915588](https://github.com/microsoft/mixed-reality-extension-unity/pull/151#issuecomment-624915588) [](ancestors = 624915588)

MREUnityRuntime/MREUnityRuntimeLib/Messaging/Protocols/Protocol.cs
eanders-MS: Unintentional. Will restore.

---
In reply to: [421119263](https://github.com/microsoft/mixed-reality-extension-unity/pull/151#discussion_r421119263) [](ancestors = 421119263)

tombuMS:
Gotcha.  Just wanted to understand the threading change.  Makes sense now.  Thanks!

---
In reply to: [625031320](https://github.com/microsoft/mixed-reality-extension-unity/pull/151#issuecomment-625031320) [](ancestors = 625031320,624915588)

MREUnityRuntime/MREUnityRuntimeLib/Messaging/Protocols/Protocol.cs
tombuMS: Ok.  Thanks!

---
In reply to: [421241477](https://github.com/microsoft/mixed-reality-extension-unity/pull/151#discussion_r421241477) [](ancestors = 421241477,421119263)

eanders-MS: The issue is there's no json here to log. It's getting deserialized within WebSocket. I have commented-out code there to reinstate message logging once another change comes in (a globally accessable logger -- WebSocket doesn't have a reference to `App`).

tombuMS: We should track this somewhere to ensure we don't forget about this logging.  It is one of the more important logging statements for debugging android issues.

---
In reply to: [428248232](https://github.com/microsoft/mixed-reality-extension-unity/pull/151#discussion_r428248232) [](ancestors = 428248232)

PR #153, opened 2020-05-08
Improved texture rendering

eanders-MS:
Textures downloaded by MREs render very aliased. This is largely due to: no mipmaps, no alpha blending. Changes in this PR:
* Generate mipmaps on downloaded textures
* Enable alpha blending in shader

### DIFF
--- a/MRETestBed/Assets/TestBed Assets/Shaders/MREDiffuseVertex.shader
+++ b/MRETestBed/Assets/TestBed Assets/Shaders/MREDiffuseVertex.shader
@@ -8,19 +8,17 @@ Shader "MRE/DiffuseVertex" {
 		_EmissiveTex ("Emissive Texture", 2D) = "white" {}
 
 		// Blend settings
-		_AlphaCutoff("Alpha Cutoff", Range(0,1)) = 0.5
-		[HideInInspector] _ShouldCutout("__should_cutout", Float) = 0.0
 		[HideInInspector] _SrcBlend("__src", Float) = 1.0
 		[HideInInspector] _DstBlend("__dst", Float) = 0.0
 		[HideInInspector] _ZWrite("__zw", Float) = 1.0
 	}
 	SubShader {
-		Tags { "RenderMode" = "Opaque" "Queue" = "Geometry" }
+		Tags { "RenderMode" = "Transparent" "Queue" = "Geometry" }
 		LOD 200
 
 		Pass {
 			Tags { "LightMode" = "Vertex" }
-			Blend [_SrcBlend] [_DstBlend]
+            Blend SrcAlpha OneMinusSrcAlpha
 			ZWrite [_ZWrite]
 
 			CGPROGRAM
@@ -38,8 +36,6 @@ Shader "MRE/DiffuseVertex" {
 
 			uniform float4 _Color;
 			uniform float4 _EmissiveColor;
-			uniform float _ShouldCutout;
-			uniform float _AlphaCutoff;
 
 			struct appdata
 			{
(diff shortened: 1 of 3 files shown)


MRETestBed/Assets/TestBed Assets/Shaders/MREDiffuseVertex.shader
```diff
@@ -8,19 +8,17 @@ Shader "MRE/DiffuseVertex" {
...
 		// Blend settings
-		_AlphaCutoff("Alpha Cutoff", Range(0,1)) = 0.5
-		[HideInInspector] _ShouldCutout("__should_cutout", Float) = 0.0
 		[HideInInspector] _SrcBlend("__src", Float) = 1.0
 		[HideInInspector] _DstBlend("__dst", Float) = 0.0
 		[HideInInspector] _ZWrite("__zw", Float) = 1.0
 	}
 	SubShader {
-		Tags { "RenderMode" = "Opaque" "Queue" = "Geometry" }
+		Tags { "RenderMode" = "Transparent" "Queue" = "Geometry" }
```
stevenvergenz, line 16: Overdraw is the death of performance on Android especially. I don't think we can enable this by default. Also, if we did want this, we should change this in the SDK and not in the shader (`material.alphaMode = MRE.AlphaMode.Blend`).

MRETestBed/Assets/TestBed Assets/Shaders/MREDiffuseVertex.shader
```diff
@@ -122,7 +118,7 @@ Shader "MRE/DiffuseVertex" {
 					fixed4(_EmissiveColor.rgb, 0.0) * tex2D(_EmissiveTex, TRANSFORM_TEX(i.uv, _EmissiveTex)) +
 					i.diffuse * tex2D(_MainTex, TRANSFORM_TEX(i.uv, _MainTex));
 				UNITY_APPLY_FOG(i.fogCoord, color);
-				clip(lerp(1, color.a - _AlphaCutoff, _ShouldCutout));
+				clip(lerp(1, color.a, 0));
```
stevenvergenz, line 121: [clip](https://developer.download.nvidia.com/cg/clip.html) discards the current pixel if the passed value is less than 0. This lerp expression always evaluates to 1, so is basically a computationally expensive no-op. Though re: my previous comment, we shouldn't be changing this shader at all for this.

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetLoader.cs
```diff
@@ -292,7 +292,7 @@ internal void OnAssetUpdate(AssetUpdate payload, Action onCompleteCallback)
 				if (!_owner) return;
 
 				var mat = asset as UnityEngine.Material;
-				var tex = asset as UnityEngine.Texture;
+				var tex = asset as UnityEngine.Texture2D;
```
stevenvergenz, line 295: If we keep this type as a `UnityEngine.Texture` we keep things open to add `RenderTexture` support later. Is there a reason this has to be a `Texture2D`?

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetLoader.cs
```diff
@@ -355,12 +355,19 @@ internal async void OnCreateAsset(CreateAsset payload, Action onCompleteCallback
...
 					response.FailureMessage = result.FailureMessage;
 				}
+				else
+				{
+					var wwwTexture = result.Asset;
+					var mipTexture = new Texture2D(wwwTexture.width, wwwTexture.height, wwwTexture.format, true, false);
+					Graphics.CopyTexture(wwwTexture, 0, 0, mipTexture, 0, 0);
+					mipTexture.Apply(true);
+					unityAsset = mipTexture;
+				}
```
stevenvergenz, line 370: Textures are ~90% of an MRE's memory footprint. I don't think we have enough margin here to increase the size of all textures by 33%. We're already short on GPU mem without this. That said, mipmaps are super useful. I think this should be exposed to the SDK, defaulting to false, so the developer can decide.

MRETestBed/Assets/TestBed Assets/Shaders/MREDiffuseVertex.shader
eanders-MS: Oh are you saying that alpha blending is already supported in the SDK? If so, these changes are completely unnecessary. Alpha blending is the high order bit here. Mipmaps are nice to have.

stevenvergenz: Yes, it's already supported. The RenderMode and everything else gets set in the MaterialPatcher.

eanders-MS:
Closing. Alpha blending already supported, turns out! We can add mipmap support later.

PR #167, opened 2020-05-23
Jabenk/mouse pick

jabenk:
- disable debug statements 
- correct and angular velocity computation


jabenk:
- as agreed with Nikola I will just PR this into physics.

PR #169, opened 2020-05-26
Testbed mrtk

tombuMS:
The scene that currently supports MRTK is the Assets/Scenes/Localhost_MRTK


eanders-MS:
Do you need to update MREUnityProjects.default.xml with the new project reference?

eanders-MS:
I opened the project in Unity and switched to the Localhost_MRTK scene, but the "MRE Component" and "Launch MRE" script references were broken until I rebuilt the dll. We'll need this to work "out of the box".

tombuMS:
> I opened the project in Unity and switched to the Localhost_MRTK scene, but the "MRE Component" and "Launch MRE" script references were broken until I rebuilt the dll. We'll need this to work "out of the box".

You will need to add them to the xml file.  I added them manually since I already had my environment set up.  I will add them to the initialization scripts that Soren added for getting started out of the box with the repo.

PR #173, opened 2020-06-01
Make asset cache write-safe

stevenvergenz:
* Rename `AssetCache` to `AssetManager`, make this local to MRE instances
* Add new class `AssetCache` that maintains immutable shared asset references and reference counts, and automatically deallocates assets 30 seconds after the reference count reaches zero
* Add ability for `AssetManager` to clone write-safe assets out of the asset cache on demand
* Add ability for client to resolve relative URLs


MREUnityRuntime/MREUnityRuntimeLib/App/IMixedRealityExtensionApp.cs
```diff
@@ -83,7 +84,7 @@ public interface IMixedRealityExtensionApp
 		/// <summary>
 		/// The url of the MRE server. Only valid after `Startup` has been called.
```
tombuMS, line 85: >url [](start = 10, length = 3)

Uri or Url?  The name of the property has now changed to Uri, so would be good if this comment aligned to the name and whether it is a url or uri.

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -107,7 +108,12 @@ private enum AppState
 		public bool IsActive => _conn?.IsActive ?? false;
 
 		/// <inheritdoc />
-		public string ServerUrl { get; private set; }
+		public Uri ServerUri { get; private set; }
+
+		/// <summary>
+		/// Same as ServerUrl, but with ws(s): substituted for http(s):
```
tombuMS, line 114: > ServerUrl [](start = 13, length = 10)

Same question here as in the interface file regarding uri versus url.

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetCache.cs
```diff
@@ -1,155 +1,188 @@
...
-				cacheCallbacks.GetOrCreate(id, () => new List<CacheCallback>(3)).Add(callback);
+				cacheItem = new CacheItem(uri, assets.ToArray(), version, assets.Count());
+				Cache.Add(uri, cacheItem);
 			}
 		}
 
-		/// <inheritdoc cref="CacheAsset"/>
-		public void CacheAsset(Object asset, Guid id, Guid containerId, AssetSource source = null, ColliderGeometry colliderGeo = null)
+		///<inheritdoc/>
+		public Task<IEnumerable<Object>> LeaseAssets(Uri uri, string ifMatchesVersion = null)
```
tombuMS, line 111: >Task<IEnumerable<Object>> [](start = 9, length = 25)

The API here all being task based is going to result in allocations of tasks just to return a value.  Though I do see merit in these being asynchronous if there was some kind of backing storage or inherent asynchronous nature to these functions, I would consider not going this route without the actual need.  Tasks come with an overhead that are not needed here.  This can always be added in the future if there becomes a need, with little to no extra work.  Just some thoughts on this from a performance perspective.  Same applies to the other task return type methods here.

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetLoader.cs
```diff
@@ -287,24 +360,21 @@ private async Task<IList<Asset>> LoadAssetsFromGLTF(AssetSource source, Guid con
...
 			{
 				if (!_owner) return;
 
-				var mat = asset as UnityEngine.Material;
-				var tex = asset as UnityEngine.Texture;
-				if (def.Material != null)
+				if (def.Material != null && asset.Asset != null && asset.Asset is UnityEngine.Material mat)
 				{
+					// make sure material reference is write-safe
+					mat = _app.AssetManager.GetById(def.Id, writeSafe: true).Value.Asset as UnityEngine.Material;
```
tombuMS, line 370: >GetById(def.Id, writeSafe: true).Value.Asset [](start = 29, length = 44)

Is this safe to dereference the value of the nullable when the GetById method can return null here and below?

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetManager.cs
```diff
@@ -0,0 +1,318 @@
...
+		/// <summary>
+		/// Break references to all shared assets and destroy all unshared assets with this container ID.
+		/// </summary>
+		/// <param name="containerId"></param>
+		public void Unload(Guid containerId)
+		{
+			var assets = Assets.Values.Where(c => c.ContainerId == containerId && c.Asset != null).ToArray();
+			for (var i = 0; i < assets.Length; i++)
+			{
+				var asset = assets[i];
```
tombuMS, line 191: This can be simplified to a ranged foreach loop since i  doesn't appear to be needed for anything other than indexing the current asset for the loop.

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -816,7 +822,7 @@ private void SetCollider(ColliderPatch colliderPatch)
 
 			if (colliderType == ColliderType.Auto)
 			{
-				colliderGeometry = App.AssetCache.GetColliderGeometry(MeshId);
+				colliderGeometry = App.AssetManager.GetById(MeshId).Value.ColliderGeometry;
```
tombuMS, line 825: >.GetById(MeshId).Value [](start = 39, length = 22)

Is this safe to just dereference when the GetBy Id can return null?

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -1043,6 +1061,14 @@ internal static void ApplyVisibilityUpdate(Actor actor, bool force = false)
 			}
 		}
 
+		private void CheckMaterialReferenceChanged(Guid id)
+		{
+			if (this != null && MaterialId == id && Renderer != null)
+			{
+				Renderer.sharedMaterial = (Material)App.AssetManager.GetById(id).Value.Asset;
```
tombuMS, line 1068: >.GetById(id).Value.Asse [](start = 56, length = 23)

Is this safe to just dereference when the GetBy Id can return null?

MREUnityRuntime/MREUnityRuntimeLib/App/IMixedRealityExtensionApp.cs
stevenvergenz: Good call, I'll update this.

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetLoader.cs
stevenvergenz: It is here, because this is inside the OnSet callback, so we are guaranteed that the asset already exists.

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
stevenvergenz: This function is an event handler that is only called with valid asset IDs, so it will never return null as it's used in this PR.

MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetCache.cs
stevenvergenz: @sorenhan agrees, I'll make these functions synchronous.

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
stevenvergenz: The `SetCollider` method is never called unless the mesh is loaded. There are guards in `PatchCollider` and `PatchAppearance` handling the loading cases.

PR #176, opened 2020-06-08
Fix object grabbing

jabenk:


MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -1367,7 +1367,7 @@ private void PatchGrabbable(bool? grabbable)
 						}
 					}
 				}
-
+				((ITargetBehavior)behaviorComponent.Behavior).Grabbable = grabbable.Value;
```
stevenvergenz, line 1370: Can we assume this is an ITargetBehavior? This actor could have a tool behavior on it, right?

stevenvergenz:
> 					if (wasGrabbable != grabbable.Value)

This is the only block that depends on a RigidBody, shouldn't that check be here? That way the previous check will pick up the non-rigidbody grab state.

---
Refers to: MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs:1358 in 70a797a. [](commit_id = 70a797a5ce5bd42c57cf4989e3d300411a902fd1, deletion_comment = False)

MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
jabenk: I think for the case when there is no rigid body it should be safe to assume (since this was also the case before physics bridge integration). Maybe then this should be then in an else branch with some additional checks (?)

jabenk: Now I separated the rigid-body case and the non-rigid body case as it was before physics bridge.

jabenk:
As I understand only the block  "if (RigidBody != null && ..." is rigid body dependent in the else branch we have the previous handling that fixes the issues the test.

jabenk:
Now only behavior that is targetable can be grabbed.

PR #179, opened 2020-06-10
Don't use physics bridge for kinematic RBs

nik0la31:


MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -289,7 +305,7 @@ internal void SynchronizeApp(ActorComponentType? subscriptionsOverride = null)
 					actorPatch.ParentId = ParentId;
 				}
 
-				if (RigidBody == null)
+				if (RigidBody == null || RigidBody.IsKinematic)
```
jabenk, line 308: !IsSimulationOwner (?)

PR #182, opened 2020-06-10
transfer ownership on grab

nik0la31:
Transfer rigid body simulation ownership when rigid body is grabbed.

Tested in both TestBed and AltspaceVR:
- when grabbed in Altspace, RBs get identity orientation


MREUnityRuntime/MREUnityRuntimeLib/Core/ActorManager.cs
```diff
@@ -188,7 +199,8 @@ internal bool OnActorDestroy(Guid id)
 
 				actor.RigidBodyAdded -= OnRigidBodyAdded;
 				actor.RigidBodyRemoved -= OnRigidBodyRemoved;
-				actor.RigidBodyKinematicsChanged -= OnActorRigidBodyKinematicsChanged;
+				actor.RigidBodyKinematicsChanged -= OnRigidBodyKinematicsChanged;
+				actor.RigidBodyOwnerChanged += OnRigidBodyOwnerChanged;
```
tombuMS, line 203: > += [](start = 31, length = 3)

I believe this should be the -= to unsubscribe like the others?

nik0la31: absolutely. thanks

PR #184, opened 2020-06-12
If there is no rigid body update do not send physics bridge update

jabenk:
- no physics bridge update when there are no rigid bodies (physics) involved.


jabenk:
https://github.com/microsoft/mixed-reality-extension-sdk/issues/610

PR #188, opened 2020-06-17
Option to disable physics bridge for backward compatibility

nik0la31:


MREUnityRuntime/MREUnityRuntimeLib/Constants.cs
```diff
@@ -54,5 +54,12 @@ static Constants()
 		 * this *is* a big deal, and requires discussion and signoff from the dev team.
 		 */
 		internal const string MinimumSupportedSDKVersion = "0.13";
+
+		/*
+		 * Enable physics bridge
+		 * Work in progress feature to improve shared physics synchronization.
+		 * It is NOT compatibile with SDK versions older than 0.19.
+		 */
+		internal const bool UsePhysicsBridge = true;
```
jabenk, line 63: I tried to use an automated expression, like string.compare with the version as an input but the compiler did not accept it

nik0la31, line 63: @tombuMS Should I put this to false right away or just in a release branch?

tombuMS: I would just have this as false in red for the moment, and you guys can enable in feature branches for your development.  Once it is ready, we can decide if this flag should become a version check for backwards compatibility or if we just remove the old system altogether and do a forced update for all SDK apps.  Would that be acceptable for you guys?

---
In reply to: [441662817](https://github.com/microsoft/mixed-reality-extension-unity/pull/188#discussion_r441662817) [](ancestors = 441662817)

nik0la31: agree, changing default setting to false right away

nik0la31: done

nik0la31: we'll figure it out by the time we enable physics bridge by default

---
In reply to: [441634208](https://github.com/microsoft/mixed-reality-extension-unity/pull/188#discussion_r441634208) [](ancestors = 441634208)

PR #196, opened 2020-06-23
Jabenk/physics dev

jabenk:
Add more safeguards to the prediction.

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
@@ -97,11 +97,20 @@ namespace MixedRealityExtension.Core
 		{
 			get
 			{
-				return Owner.HasValue ? Owner.Value == App.LocalUser.Id : CanSync();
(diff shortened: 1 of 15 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -875,6 +875,7 @@ private RigidBody AddRigidBody()
 			if (_rigidbody == null)
 			{
 				_rigidbody = gameObject.AddComponent<Rigidbody>();
+				//_rigidbody.maxDepenetrationVelocity = 3.0F; // set very low de-penetration velocities
```
nik0la31, line 878: >//_rigidbody.maxDepenetrationVelocity = 3.0F; // set very low de-penetration velocities [](start = 4, length = 87)

should we uncomment this?

jabenk: For the moment I would leave it as a comment because depenetration helps when one remote body pushes one owned body or vice versa and even without this limit we have quite some good behavior.

PR #197, opened 2020-06-23
Add in our link.xml file to be copied to build output directory

tombuMS:
This has also been updated on the build server to copy to output to the artifacts directory for distribution.

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/link.xml
+++ b/MREUnityRuntime/MREUnityRuntimeLib/link.xml
@@ -1,0 +1,6 @@
+<linker>
+  <assembly fullname="MREUnityRuntimeLib" preserve="all"/>
+  <assembly fullname="System">
+    <type fullname="Mono.Net.Security.LegacyTlsProvider" preserve="all" />
+  </assembly>
+</linker>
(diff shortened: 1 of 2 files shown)


tombuMS:
Am in the process of adding the copying to the post build tool and rebuilding it.

tombuMS:
Do we think that we want to add this to the PostBuildTool?  Not sure if we need to copy this over each time we do a build.  It doesn't get updated hardly at all.  Do we want to just have it as our checked in file to the test beds, and as a part of the build artifacts that are generated by the build server to include in our updates to host apps?

MREUnityRuntime/MREUnityRuntimeLib/link.xml
```diff
@@ -0,0 +1,6 @@
+<linker>
+  <assembly fullname="MREUnityRuntimeLib" preserve="all"/>
+  <assembly fullname="System">
+    <type fullname="Mono.Net.Security.LegacyTlsProvider" preserve="all" />
+  </assembly>
+</linker>
```
stevenvergenz, line 6: Is this all? I thought the MRE runtime required System.Net included as well.

tombuMS: This is copied directly from the Test Bed link.xml file.  I would have thought this would be all inclusive of what we need to run.

tombuMS: I also do not see the one you mention in the Altspace link.xml file either.

stevenvergenz: I confused csc.rsp with link.xml. This is only needed by the compiler, not the linker. Approved.

PR #201, opened 2020-06-25
Jabenk/physics dev

jabenk:
- minor improvements

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Patching/PatchingUtils.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Patching/PatchingUtils.cs
@@ -211,6 +211,8 @@ namespace MixedRealityExtension.Patching
 			var patch = new RigidBodyPatch()
 			{
 				// Do not include Position or Rotation in the patch.
+				Velocity = GeneratePatch(_old.Velocity, sceneRoot.InverseTransformDirection(_new.velocity)),
+				AngularVelocity = GeneratePatch(_old.AngularVelocity, sceneRoot.InverseTransformDirection(_new.angularVelocity)),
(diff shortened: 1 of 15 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Patching/PatchingUtils.cs
```diff
@@ -212,9 +212,10 @@ public static RigidBodyPatch GeneratePatch(RigidBody _old, Rigidbody _new, Trans
 			{
 				// Do not include Position or Rotation in the patch.
 
-				// <todo> 
-				//Velocity = GeneratePatch(_old.Velocity, sceneRoot.InverseTransformDirection(_new.velocity)),
-				//AngularVelocity = GeneratePatch(_old.AngularVelocity, sceneRoot.InverseTransformDirection(_new.angularVelocity)),
+				Velocity = ((addVelocities) ?
+				  GeneratePatch(_old.Velocity, sceneRoot.InverseTransformDirection(_new.velocity)) : null),
+				AngularVelocity = ((addVelocities) ?
+				  GeneratePatch(_old.AngularVelocity, sceneRoot.InverseTransformDirection(_new.angularVelocity)) : null),
```
nik0la31, line 218: may be a bit more readable if you wrap twoof these with an if

jabenk: I am not a C# expert but I think in such field initialization code you cannot add IF. I tried and got error, but maybe there are other ways. 
This way with the tenor works for sure. 

---
In reply to: [445599978](https://github.com/microsoft/mixed-reality-extension-unity/pull/201#discussion_r445599978) [](ancestors = 445599978)

PR #206, opened 2020-07-01
Jabenk/smallchanges

jabenk:
Add a separate subscription for the velocities of the rigid body. 
https://github.com/microsoft/mixed-reality-extension-sdk/issues/610

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
@@ -1670,7 +1670,7 @@ namespace MixedRealityExtension.Core
 			}
 
 			// If the actor has a rigid body then always sync the transform and the rigid body.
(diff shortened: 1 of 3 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Core/Actor.cs
```diff
@@ -1646,6 +1650,7 @@ private bool ShouldSync(ActorComponentType subscriptions, ActorComponentType fla
 			}
 
 			// If the actor has a rigid body then always sync the transform and the rigid body.
+			// but not the velocity, only
```
nik0la31, line 1653: >// but not the velocity, only  [](start = 3, length = 30)

something is missing?

jabenk: fixed

PR #216, opened 2020-07-10
User ID Generation Update

tombuMS:
The User ID obfuscation now happens in the MRE Unity Runtime and is based of of either a combination of the host supplied user ID and the app ID, or a random generated GUID.  This establishes that for user tracking to be truly persistent, that an app much be registers and have an app ID.


MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -783,6 +780,33 @@ bool ancestorInList(GameObject go)
 			}
 		}
 
+		private Guid GenerateObfuscatedUserId(IHostAppUser hostAppUser)
+		{
+			if (GrantedPermissions.HasFlag(Permissions.UserTracking) && GlobalAppId != string.Empty)
```
stevenvergenz, line 785: Would be good to include a null test here too.

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -783,6 +780,33 @@ bool ancestorInList(GameObject go)
...
+			{
+				using (SHA256 hasher = SHA256.Create())
+				{
+					var encoder = new UTF8Encoding();
+
+					StringBuilder idString = new StringBuilder();
```
stevenvergenz, line 791: If it's just the one interpolated string, why use a string builder?

tombuMS: This was originally appended to later when I was working through this implementation.  Can just be a string now and I will update it to be.

---
In reply to: [453088797](https://github.com/microsoft/mixed-reality-extension-unity/pull/216#discussion_r453088797) [](ancestors = 453088797)

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
tombuMS: We do guarantee that is not null at the time that the constructor is called, just so that we don't have to null check this value everywhere it is read from.

---
In reply to: [453088588](https://github.com/microsoft/mixed-reality-extension-unity/pull/216#discussion_r453088588) [](ancestors = 453088588)

PR #219, opened 2020-07-14
Keep last packet at least for 100ms for possible interpolation

nik0la31:


jabenk:
I guess this fixes now the effects on the video that rarely appear.

PR #223, opened 2020-07-17
Initialize asset cache earlier

stevenvergenz:
By moving the init to `Awake()` instead of `Start()`, MREs can be connected in `Start()` without causing problems.


tombuMS:
> 			Application.lowMemory += CleanUnusedResources;

Not a part of these changes, but I noticed that this event is not unsubscribed from in the asset cache in the case that it is destroyed.  We should add that as this will cause a memory leak whenever the asset cache is destroyed and the host application lives on.

---
Refers to: MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetCache.cs:90 in 26097cd. [](commit_id = 26097cd71f6512ca785896a23f6f0d6172553459, deletion_comment = False)

PR #225, opened 2020-08-04
Change assumptions to 'UserJoin' all users, not just local ones

stevenvergenz:
* In #216 we accidentally removed the ability to attach things to anyone but the local user. We removed the `UserInfoProvider`, and replaced it with an `IHostAppUser` that provides the attach points. These host app users are provided to the runtime through `IMixedRealityApp.UserJoin`.
* I have added an `isLocalUser` boolean argument to that method so host app user objects for non-local users can also be passed in. This changes the convention of only local users calling UserJoin; now UserJoin should be called for all users that the client knows about.
* For non-local users on a client, we now compute both the AppId-based user ID and also the EphemeralAppId-based ID for the purposes of looking up attachments, since we don't know which one the user is going by.

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/API/MREApi.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/API/MREApi.cs
@@ -138,10 +138,19 @@ namespace MixedRealityExtension.API
 		/// Creates a new mixed reality extension app and adds it to the MRE runtime.
 		/// </summary>
 		/// <param name="ownerScript">The owner unity script for the app.</param>
-		/// <param name="globalAppId">The global app id for the app being instanced, or empty if there is no global app id.</param>
-		/// <param name="ephemeralAppId">A string uniquely identifying the MRE instance on all clients.</param>
+		/// <param name="ephemeralAppId">A unique identifier for the MRE behind this instance's URL, in the absence
+		/// of a global app ID. Used for generating user IDs that are consistent within this session across clients,
+		/// but not reliable across time. Must be synchronized across all clients in this session, and must be
+		/// periodically rotated.</param>
+		/// <param name="globalAppId">A unique identifier for the MRE behind this instance's URL. Used for generating
+		/// consistent user IDs for this MRE. Would typically come from an app registry or similar. If supplied, must
+		/// be synchronized across all clients in this session.
+		/// </param>
 		/// <returns>Returns the newly created mixed reality extension app.</returns>
-		public IMixedRealityExtensionApp CreateMixedRealityExtensionApp(MonoBehaviour ownerScript, string ephemeralAppId, string globalAppId = "")
+		public IMixedRealityExtensionApp CreateMixedRealityExtensionApp(
+			MonoBehaviour ownerScript,
+			string ephemeralAppId,
+			string globalAppId = "")
 		{
 			var mreApp = new MixedRealityExtensionApp(globalAppId ?? string.Empty, ephemeralAppId, ownerScript)
 			{
--- a/MREUnityRuntime/MREUnityRuntimeLib/API/MREApi.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/API/MREApi.cs
@@ -139,11 +139,11 @@ namespace MixedRealityExtension.API
 		/// </summary>
 		/// <param name="ownerScript">The owner unity script for the app.</param>
 		/// <param name="globalAppId">The global app id for the app being instanced, or empty if there is no global app id.</param>
-		/// <param name="localAppId">A string uniquely identifying the MRE instance on all clients.</param>
+		/// <param name="ephemeralAppId">A string uniquely identifying the MRE instance on all clients.</param>
 		/// <returns>Returns the newly created mixed reality extension app.</returns>
-		public IMixedRealityExtensionApp CreateMixedRealityExtensionApp(MonoBehaviour ownerScript, string globalAppId = "", string localAppId = "")
+		public IMixedRealityExtensionApp CreateMixedRealityExtensionApp(MonoBehaviour ownerScript, string ephemeralAppId, string globalAppId = "")
 		{
-			var mreApp = new MixedRealityExtensionApp(globalAppId ?? string.Empty, localAppId, ownerScript)
+			var mreApp = new MixedRealityExtensionApp(globalAppId ?? string.Empty, ephemeralAppId, ownerScript)
 			{
 				InstanceId = Guid.NewGuid()
 			};
--- a/MREUnityRuntime/MREUnityRuntimeLib/API/MREApi.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/API/MREApi.cs
@@ -139,10 +139,11 @@ namespace MixedRealityExtension.API
 		/// </summary>
 		/// <param name="ownerScript">The owner unity script for the app.</param>
 		/// <param name="globalAppId">The global app id for the app being instanced, or empty if there is no global app id.</param>
+		/// <param name="localAppId">A string uniquely identifying the MRE instance on all clients.</param>
 		/// <returns>Returns the newly created mixed reality extension app.</returns>
-		public IMixedRealityExtensionApp CreateMixedRealityExtensionApp(MonoBehaviour ownerScript, string globalAppId = "")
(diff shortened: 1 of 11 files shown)


MREUnityRuntime/MREUnityRuntimeLib/API/MREApi.cs
```diff
@@ -139,10 +139,11 @@ public class MREAppsAPI
 		/// </summary>
 		/// <param name="ownerScript">The owner unity script for the app.</param>
 		/// <param name="globalAppId">The global app id for the app being instanced, or empty if there is no global app id.</param>
+		/// <param name="ephemeralAppId">A string uniquely identifying the MRE instance on all clients.</param>
 		/// <returns>Returns the newly created mixed reality extension app.</returns>
```
sorenhan, line 143: While this is going to make the description longer, would like to change to clarify that this value should be identical across all clients talking together, but should ideally change at regular intervals (for example, if host app has a multiplayer server instance that synchronize users, then instance unique ID)?) Clarify that it's important it changes to support users that decline user tracking)?

Maybe for globalAppID, add in that the global app id should be persistent

sorenhan:
> 		}

remove instanceduserid also

---
Refers to: MREUnityRuntime/MREUnityRuntimeLib/Core/UserManager.cs:32 in b283a3b. [](commit_id = b283a3b71c5d2737a860770f7fd5cca1e6f1fc8e, deletion_comment = False)

MREUnityRuntime/MREUnityRuntimeLib/Core/User.cs
```diff
@@ -23,10 +23,13 @@ internal class User : MixedRealityExtensionObject, IUser
...
 		public UInt32 Groups { get; internal set; } = 1;
 
-		internal void Initialize(IHostAppUser hostAppUser, Guid userId, MixedRealityExtensionApp app)
+		public Guid InstancedUserId { get; private set; }
+
+		internal void Initialize(IHostAppUser hostAppUser, Guid userId, Guid instancedUserId, MixedRealityExtensionApp app)
 		{
 			HostAppUser = hostAppUser;
 			base.Initialize(userId, app);
+			InstancedUserId = instancedUserId;
```
sorenhan, line 32: >InstancedUserId [](start = 3, length = 15)

maybe rename instancedUserID to EphemeralUserId throughout the code?

sorenhan:
> 		}

Shouldn't all of this be inside the if(isLocalUser)? I'd think the only thing that should be done for non-local users is the adduser/removeuser calls?

---
Refers to: MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs:526 in b283a3b. [](commit_id = b283a3b71c5d2737a860770f7fd5cca1e6f1fc8e, deletion_comment = False)

sorenhan:
> 				OnUserLeft?.Invoke(user);

same for userJoin - everything except _userManager.RemoveUser() should be in an if(islocaluser) check

---
Refers to: MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs:549 in b283a3b. [](commit_id = b283a3b71c5d2737a860770f7fd5cca1e6f1fc8e, deletion_comment = False)

stevenvergenz:
> 		}

Right now, the only thing that's happening for non-locals is the addition of a User monobehaviour, the generation of IDs, and the addUser call. We need all of these, don't we?

---
In reply to: [670192762](https://github.com/microsoft/mixed-reality-extension-unity/pull/225#issuecomment-670192762) [](ancestors = 670192762)

---
Refers to: MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs:526 in b283a3b. [](commit_id = b283a3b71c5d2737a860770f7fd5cca1e6f1fc8e, deletion_comment = False)

MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs
```diff
@@ -474,27 +481,36 @@ void PerformUserJoin()
 				if (user == null)
 				{
 					user = userGO.AddComponent<User>();
+
 					// Generate the obfuscated user ID based on user tracking permission.
-					user.Initialize(hostAppUser, GenerateObfuscatedUserId(hostAppUser), this);
+					Guid instancedUserId = GenerateObfuscatedUserId(hostAppUser, EphemeralAppId);
```
tombuMS, line 486: So is the plan to push the randomization of a user who has denied user tracking to the host app again?

MREUnityRuntime/MREUnityRuntimeLib/Core/User.cs
tombuMS: I agree in keeping the ephemeral naming convention consistent here as well.

stevenvergenz:
> 				OnUserLeft?.Invoke(user);

Ah yes, I missed the UserLeft function. Good catch.

---
In reply to: [670193108](https://github.com/microsoft/mixed-reality-extension-unity/pull/225#issuecomment-670193108) [](ancestors = 670193108)

---
Refers to: MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs:549 in b283a3b. [](commit_id = b283a3b71c5d2737a860770f7fd5cca1e6f1fc8e, deletion_comment = False)

sorenhan:
> 		}

Is there a concern if user.Id == user.EphemeralUserId? Should we have a check to only add/remove the instanceduserid/ephemeraluserid if they are not identical?

---
In reply to: [670192440](https://github.com/Microsoft/mixed-reality-extension-unity/pull/225#issuecomment-670192440) [](ancestors = 670192440)

---
Refers to: MREUnityRuntime/MREUnityRuntimeLib/Core/UserManager.cs:32 in b283a3b. [](commit_id = b283a3b71c5d2737a860770f7fd5cca1e6f1fc8e, deletion_comment = False)

tombuMS:
Here are my major questions about the design of this, as I don't have access to codeflow at the moment.

1. Are we now making the random seed a part of the host app’s responsibility now?  
2. Should we be calling OnUserJoined for non-local users?  
3. Do we need to add non-local users to the user manager?  Is that how you are trying to handle the mapping for attachments?

sorenhan:
> 		}

sorry, I misread the code here, you're right about most of this. what about OnUserJoined?.Invoke(user)? I don't know what's relying on this, and if there are effects when this is called for every user now.

---
In reply to: [670195274](https://github.com/Microsoft/mixed-reality-extension-unity/pull/225#issuecomment-670195274) [](ancestors = 670195274,670192762)

---
Refers to: MREUnityRuntime/MREUnityRuntimeLib/App/MixedRealityExtensionApp.cs:526 in b283a3b. [](commit_id = b283a3b71c5d2737a860770f7fd5cca1e6f1fc8e, deletion_comment = False)

stevenvergenz:
> 		}

If they are identical, then this is a no-op. Seems redundant to check it.

---
In reply to: [670200200](https://github.com/microsoft/mixed-reality-extension-unity/pull/225#issuecomment-670200200) [](ancestors = 670200200,670192440)

---
Refers to: MREUnityRuntime/MREUnityRuntimeLib/Core/UserManager.cs:32 in b283a3b. [](commit_id = b283a3b71c5d2737a860770f7fd5cca1e6f1fc8e, deletion_comment = False)

MREUnityRuntime/MREUnityRuntimeLib/API/MREApi.cs
tombuMS: Agree that the documentation header should include the necessity of the ID being rotated or ephemeral in the use case of a user denying tracking.

MREUnityRuntime/MREUnityRuntimeLib/API/MREApi.cs
```diff
@@ -139,10 +139,11 @@ public class MREAppsAPI
...
-		public IMixedRealityExtensionApp CreateMixedRealityExtensionApp(MonoBehaviour ownerScript, string globalAppId = "")
+		public IMixedRealityExtensionApp CreateMixedRealityExtensionApp(MonoBehaviour ownerScript, string ephemeralAppId, string globalAppId = "")
```
tombuMS: The default value for the global app ID should be removed now that we have these two ID's. Both are really required in some kind of fashion if a host app wants to allow any kind of user tracking.

PR #229, opened 2020-08-19
Update write-safe materials on prefabs, even if prefab has multiple mats

stevenvergenz:


MRETestBed/Assets/TestBed Assets/Scripts/MREComponent.cs
```diff
@@ -230,13 +230,13 @@ private void MREApp_OnConnectFailed(MixedRealityExtension.IPC.ConnectFailedReaso
 		}
 	}
 
-	private void MRE_OnUserJoined(IUser user)
+	private void MRE_OnUserJoined(IUser user, bool isLocalUser)
```
tombuMS, line 233: >, bool isLocalUser [](start = 41, length = 18)

Is this just the result of the test bed not being updated from a previous change?  Only because it seems unrelated to this pull request.

stevenvergenz: Correct, we just needed this to get it to compile.

---
In reply to: [473272617](https://github.com/microsoft/mixed-reality-extension-unity/pull/229#discussion_r473272617) [](ancestors = 473272617)

kahunamoore:
Does this need to be merged to master? I'm unfamiliar with the branching/merging strategy used here.

FYI: I ran across compile erros (in MREComponent.cs) on master due to the added bool parameter on the OnUserJoined/OnUserLeft methods.

tombuMS:
> Does this need to be merged to master? I'm unfamiliar with the branching/merging strategy used here.
> 
> FYI: I ran across compile erros (in MREComponent.cs) on master due to the added bool parameter on the OnUserJoined/OnUserLeft methods.

This PR cannot be merged to master yet as it has not shipped.  I will make a targeted fix and get it in to the test bed.  If you want to get running until then, you can either replicate the changes (they are minor) in to the MREComponent locally on your machine, or can pull the red branch which is only a few commits ahead of master.

sorenhan:
By the way, the branching strategy is described here: https://github.com/microsoft/mixed-reality-extension-sdk/blob/master/DEVELOPER_WORKFLOW.md
It's a bit complex because of there are multiple interdependent repos for client runtimes, SDK itself, and the sample repo. Plus it is planned towards to supporting multiple live versions simultaneously.
But in general, changes go into the red or green branches. We update the master branch only after we have complete a test pass and shipped the runtime in an Altspace client update.

PR #231, opened 2020-09-02
Fix exceptions if physics sim runs before local user joins

stevenvergenz:
Just something I noticed after seeing intermittent exceptions from this area. Better safe than sorry.


tombuMS (approved):
:shipit:

PR #236, opened 2020-09-16
Fix for deadlock on cleanup in Mac OS host apps

tombuMS:
Fix an issue during cleanup of the websocket where on Mac OS, the close output only on the websocket does not cause the cancellation of the awaiting read task.


stevenvergenz (approved):
Nice cleanup!

PR #237, opened 2020-10-01
When spawning prefabs, re-assign write-safe materials

stevenvergenz:
* Write-safe assets now contain a reference to the read-only versions
* When fetching an asset by Unity object, it now checks against both read-only and write-safe versions
* Logic to listen for asset updates is now triggered on actor creation too, not just on appearance patch

Fixes microsoft/mixed-reality-extension-sdk#721


tombuMS (approved):
:shipit:

PR #238, opened 2020-10-05
Fix asset refcounting issues when server doesn't provide etags

stevenvergenz:
If the asset server is not providing `ETag` headers with its responses, assume cached assets are always fresh.


tombuMS (approved):
:shipit:

PR #239, opened 2020-10-08
Fix library materials

stevenvergenz:
Don't assign default material to new actors when current is not found in cache.


tombuMS (approved):
:shipit:

PR #240, opened 2020-10-15
Fixed a null deref when last response is null

in2dair:
Absent this change, some types of failures are masked by a null pointer deref

### DIFF
--- a/MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetLoader.cs
+++ b/MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetLoader.cs
@@ -192,7 +192,8 @@ namespace MixedRealityExtension.Assets
 				}
 				catch (HttpRequestException)
 				{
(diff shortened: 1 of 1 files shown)


MREUnityRuntime/MREUnityRuntimeLib/Assets/AssetLoader.cs
```diff
@@ -192,7 +192,8 @@ private async Task<IList<Asset>> LoadAssetsFromGLTF(AssetSource source, Guid con
 				}
 				catch (HttpRequestException)
 				{
-					if (loader.LastResponse.StatusCode == System.Net.HttpStatusCode.NotModified)
+					if (loader.LastResponse != null
+						&& loader.LastResponse.StatusCode == System.Net.HttpStatusCode.NotModified)
```
stevenvergenz, line 196: I would have used the ? operator, since it's a little shorter, but this works.

stevenvergenz (approved):

PR #241, opened 2020-10-26
Add new app events OnWaitingForPermission and OnPermissionDenied

stevenvergenz:
Also: update the permission requests so that if an app doesn't supply a manifest, the runtime assumes it requires both tracking and interaction.


tombuMS (approved):
:shipit:

PR #242, opened 2020-11-05
Add cancellation token to IPermissionManager

stevenvergenz:
Cancel permission requests if the MRE shuts down.


tombuMS (approved):
:shipit:

PR #254, opened 2021-11-02
Switch to Unity's version of Newtonsoft.Json

la-thorn:
Switching to Unity's version of the Newtonsoft.Json package makes it easier to avoid package collisions when using multiple packages from the Unity Package Manager which depend upon this common JSON library.


stevenvergenz (approved):
Wow, that easy? Is this tested yet?

la-thorn:
> Wow, that easy? Is this tested yet?

Haha yeah turns out to be that easy. Tested it on AltQuiz and the test app, seems to work just fine.

tombuMS (approved):

